PR #6, opened 2026-04-12
fix: make Bun installs use the isolated linker

paul-bouzian:
# Summary
- Make the Bun install layout explicit for this monorepo instead of relying on Bun version defaults
- Prevent workspace `prepare` scripts from failing when `@effect/language-service` expects a workspace-local `./node_modules/typescript`
- Keep the existing Bun 1.3.9 project expectation while making installs more robust for contributors on older or migrated setups

# Changes
- Add a root `bunfig.toml` with `install.linker = "isolated"`
- Align the workspace install layout with what the existing `effect-language-service patch` prepare scripts already assume
- Avoid the `scripts/node_modules/typescript/package.json` lookup failure during `bun install`

# Testing
- `bun install` - Pass
- `bun run fmt` - Pass
- `bun run lint` - Pass with existing repository warnings and no errors
- `bun run typecheck` - Pass

# Notes
- The repository already declares `bun@1.3.9` in `package.json` and `.mise.toml`; this change makes the required monorepo install layout explicit instead of relying on Bun defaults

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,0 +1,2 @@
+[install]
+linker = "isolated"


kartikkabadi:
Devin Review – PR #6 findings (0 Vulns, 1 Bug, 0 Flags, Checks 5/5):

**Bug – Clean leaves script dependencies installed (bunfig.toml:2)**
With linker = "isolated", bun run clean leaves scripts/node_modules intact. The clean script omits that workspace, so stale script dependencies survive cleanup. Isolated linking creates a separate node_modules directory for every workspace that declares dependencies. The scripts directory is a workspace and declares dependencies, so installs now create scripts/node_modules. The root clean script removes the root, app, and package dependency directories, but not this new directory. Example: Run bun install, then bun run clean. scripts/node_modules/typescript remains available even though the equivalent dependency trees under apps/* and packages/* were deleted. Please add scripts/node_modules to the root clean command alongside apps/*/node_modules and packages/*/node_modules.

Review: https://app.devin.ai/review/Emanuele-web04/synara/pull/6 – PR #6

PR #7, opened 2026-04-12
fix: repair legacy desktop upgrade migration and replay crashes

paul-bouzian:
## Why
- Desktop upgrades can fail when an existing `effect_sql_migrations` table already contains different migration names for ids `17` through `22` from an older release line.
- In that state, the current thread metadata migrations are skipped, `023_ProjectionThreadsAssociatedWorktreeRef` still runs, and startup crashes because `associated_worktree_branch` does not exist.
- Some legacy desktop databases also contain `thread.archived` or `thread.unarchived` orchestration events, which the current release cannot decode during replay, causing the backend to restart in a loop even after the schema issue is repaired.

## How
- Make migration `23` repair the full current thread/message metadata shape before backfilling `associated_worktree_ref`.
- Detect missing columns from the current `17`-`23` migration line and add only the ones that are absent, so healthy databases keep their existing data untouched.
- Add migration `24` to normalize legacy archived/unarchived thread events into decodable `thread.meta-updated` rows that preserve `threadId` and `updatedAt`.
- Add regression tests for both the migration-ledger mismatch and the legacy archived-event replay path.

## Tests
- `bunx vitest run apps/server/src/persistence/Migrations/023_ProjectionThreadsAssociatedWorktreeRef.test.ts apps/server/src/persistence/Migrations/024_NormalizeLegacyArchivedThreadEvents.test.ts`
- `bun fmt`
- `bun lint` (passes with existing repo warnings only)
- `bun typecheck`

## Review Setup
- No extra setup is required beyond the normal app deploy.
- To manually verify the upgrade path, use a desktop profile whose `effect_sql_migrations` table already contains the legacy `17`-`22` rows and whose `orchestration_events` table may include legacy `thread.archived` rows, then launch the app and confirm startup no longer crashes during replay.

### DIFF
--- a/apps/server/src/persistence/Migrations.ts
+++ b/apps/server/src/persistence/Migrations.ts
@@ -36,6 +36,7 @@ import Migration0020 from "./Migrations/020_ProjectionThreadsForkSource.ts";
 import Migration0021 from "./Migrations/021_ProjectionThreadsAssociatedWorktree.ts";
 import Migration0022 from "./Migrations/022_ProjectionThreadsAssociatedWorktreeBranch.ts";
 import Migration0023 from "./Migrations/023_ProjectionThreadsAssociatedWorktreeRef.ts";
+import Migration0024 from "./Migrations/024_NormalizeLegacyArchivedThreadEvents.ts";
 
 /**
  * Migration loader with all migrations defined inline.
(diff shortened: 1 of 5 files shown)


kartikkabadi:
Devin Review – PR #7 findings (0 Vulns, 1 Bug, 0 Flags, Checks 5/5):

**Bug – Migration 23 skips newer legacy ledgers (023_ProjectionThreadsAssociatedWorktreeRef.ts:11, R11-12)**
When a legacy ledger records ID 23, ProjectionThreadsAssociatedWorktreeRef never runs its repair. Migration 24 advances the ledger with required columns absent, so startup queries crash. Please guard repair on column presence, not just migration ID, so newer legacy ledgers still get repaired.

Review: https://app.devin.ai/review – PR #7

PR #33, opened 2026-04-19
Improve GPT model display names

Marve10s:
## Summary
- reuse shared GPT display-name formatting for custom and dynamic model options
- keep the model picker, app settings, and subagent presentation consistent
- preserve the provider settings shape needed by the current web UI while applying the naming updates

## Verification
- bun fmt
- bun lint
- bun typecheck


Marve10s:
Align gpt 5.1 namings with other GPT models
image

PR #36, opened 2026-04-20
[codex] Fix Bun startup for local history migration

Marve10s:
## Summary
- avoid importing `node:sqlite` at module load time under Bun
- open the history snapshot database with `bun:sqlite` on Bun and `node:sqlite` elsewhere
- keep legacy home migration working for localhost startup under both runtimes

## Root cause
The server imported `node:sqlite` eagerly from `apps/server/src/homeMigration.ts`. Under Bun dev startup, that built-in module is unavailable, so the server crashed before the app could boot.

## Impact
This restores local dev startup on Bun when home migration code is present, without changing the non-Bun path.

## Validation
- `bun run --cwd apps/server test src/homeMigration.test.ts`
- `bun fmt`
- `bun lint`
- `bun typecheck`

### DIFF
--- a/apps/server/src/homeMigration.ts
+++ b/apps/server/src/homeMigration.ts
@@ -2,10 +2,8 @@
  * FILE: homeMigration.ts
  * Purpose: Imports legacy ~/.t3 state into the new ~/.dpcode home on first startup.
  * Layer: Startup utility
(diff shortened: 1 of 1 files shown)


Marve10s:
Closing this because it was opened from the Bun startup verification fix instead of the requested t3code#1858 port. Replacing it with the changed-files persistence PR.

PR #37, opened 2026-04-20
Persist changed-files expansion state per thread

Marve10s:
## What changed
- Persist changed-files tree expansion state per thread instead of keeping it inside `MessagesTimeline` component state.
- Move the expansion source of truth into the web store, keyed by scoped thread id and turn id.
- Clean up stored expansion overrides when threads disappear.

## Why
The changed-files tree reset to expanded whenever the thread view unmounted and mounted again. That happened because the expansion map lived inside `MessagesTimeline` as local React state.

## Notes
- Default behavior stays the same: changed-files trees start expanded.
- Only collapsed overrides are stored; expanding again removes the override.
- This adapts the upstream `t3code#1858` behavior to DPcode's current store structure.

## Validation
- `bun fmt`
- `bun lint`
- `bun typecheck`
- `bun run --cwd apps/web test --run src/store.test.ts src/components/chat/MessagesTimeline.test.tsx`


kartikkabadi:
Devin Review – PR #37 findings (0 Vulns, 1 Bug, 0 Flags, Checks 5/5):

**Bug – Shell snapshots retain deleted threads (store.ts:3201, R3201-3204)**
When a shell snapshot omits a deleted thread, threadChangedFilesExpandedByTurnId retains its override. syncServerShellSnapshot preserves the entire map, so deletions accumulate unreachable persisted entries. Please prune overrides for threads missing from snapshot, same as syncServerReadModel does.

Review: https://app.devin.ai/review – PR #37

PR #40, opened 2026-04-20
Fix T3 import reliability and surface provider usage

sbbedefended:
## Summary
- make legacy T3 import reliable by importing full thread state in bulk and preserving thread/message metadata
- stop implicit startup migration from `~/.t3` so import only happens through explicit DP Code actions
- surface real provider usage in the UI, including the branch toolbar panel and optional composer badge

## Details
- add `orchestration.importLegacyT3State` plus the server-side importer and schema repair migration for legacy projection databases
- extend orchestration contracts, projector/read-model, projection pipeline, websocket routing, and web store to support bulk imported messages / activities / proposed plans
- add explicit T3 import entry points in the sidebar and settings through a shared `useLegacyT3Import` flow
- fix desktop confirm usage so the import prompt uses the native bridge instead of falling back incorrectly in packaged builds
- add provider usage snapshots for Codex and Claude from local CLI/session data, then reuse the same summary in `Rate limits remaining` and the optional composer badge

## Validation
- `bun fmt`
- `bun lint`
- `bun typecheck`

### DIFF
--- a/apps/server/src/legacyStateImport.ts
+++ b/apps/server/src/legacyStateImport.ts
@@ -1,0 +1,1136 @@
+import { DatabaseSync } from "node:sqlite";
+
+import {
+  CheckpointRef,
+  ChatAttachment,
+  CommandId,
+  EventId,
+  MessageId,
+  ModelSelection,
+  NonNegativeInt,
+  OrchestrationMessageRole,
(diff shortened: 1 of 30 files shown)


Emanuele-web04:
@codex resolve the merge conflicts in this pull request

Issue #43, opened 2026-04-20
[Bug]: Project add can silently succeed, then the second add fails with duplicate workspace root


justelson:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Install the desktop app fresh on Windows.
2. Open DP Code.
3. Add a local project/folder.
4. The UI shows the add flow, but no project/thread appears.
5. Try adding the same folder again.
6. The second attempt fails with a duplicate workspace-root error.

### Expected behavior

The first add should surface the project and let me start/use threads immediately. If the project already exists, the app should recover or focus that existing project instead of throwing an invariant error.

### Actual behavior

The first `project.create` appears to succeed, but the client never surfaces the project/thread. The second add then fails with:

`Orchestration command invariant failed (project.create): Project '<id>' already uses workspace root '<path>'`

### Impact

Blocks work completely

### Version or commit

Desktop app build on Apr 21, 2026

### Environment

Windows desktop app, first install / fresh local state

### Logs or stack traces

```shell
`Error: Orchestration command invariant failed (project.create): Project 'de518ba1-43c2-46ae-80af-2cbb219ad231' already uses workspace root 'C:\Users\elson\my_coding_play\play projects\pi'.`
```

### Screenshots, recordings, or supporting files

[image.png](https://github.com/user-attachments/assets/d6214498-8e9c-4b54-b50b-6130e68d7a9c)

### Workaround

No reliable user-facing workaround found. Restarting/reloading may resync state sometimes, but the add flow itself is broken.

justelson:
Fixed in #44.

Root cause: `project.create` could succeed, but the client only polled `getSnapshot()` and could accept broken local/read-model state where the created project row was still missing. That left the UI empty after the first add, and retrying the same folder then hit the duplicate workspace-root invariant.

Fix: add-project and duplicate-create flows now share the same recovery helper, which first retries the read model and then falls back to `repairState()` if the project still cannot be surfaced. If the project is recovered, the client syncs the repaired snapshot and reopens the existing project instead of failing on the second add.

PR: https://github.com/Emanuele-web04/dpcode/pull/44

cmdr-chara:
[...]

Issue #35, opened 2026-04-20
[Bug]: Threads not working


Battzz:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Yesterday i used the application and was working all good (I've added a few threads)
2. Open the app today 
3. See i have no threads
4. I try to add a new thread (I see adding.... but then nothing comes up)
5. Can't add nothing and getting error if i add the same thread again

### Expected behavior

Persisten thread and add threads working

### Actual behavior

Nothing comes up

### Impact

Blocks work completely

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[Screenshot 2026-04-20 at 08.42.52.png](https://github.com/user-attachments/assets/14f239ec-f7ae-443d-8486-d6688cd6e9e4)
[Screenshot 2026-04-20 at 08.43.24.png](https://github.com/user-attachments/assets/90238c91-18ab-40c1-94fa-272ada97753c)

Battzz:
Quick fix solution: delete .t3 and .dpcode. Seems to be a problem with migration from t3 code!

Emanuele-web04:
Thank you so much will look deeper into this

justelson:
Fixed in #44.

Root cause: the desktop client could accept a broken local/read-model snapshot where live threads still existed but their live project rows were missing. In that state the sidebar came up empty, the first add/create flow could appear to do nothing, and retrying the same folder could then hit the duplicate workspace-root invariant.

Fix: desktop bootstrap now treats “live threads with missing projects” as a repair case instead of accepting that snapshot, and the add-project / duplicate-create recovery paths now share the same recovery helper with a `repairState()` fallback. If polling still cannot surface the created project, the client repairs state and reopens the recovered project instead of leaving the UI empty.

PR: https://github.com/Emanuele-web04/dpcode/pull/44

cmdr-chara:
@Emanuele-web04 — closure candidate: the desktop bootstrap recovery for this failure is present on current `main`. `desktopProjectRecovery.ts` detects/reconciles broken desktop snapshots where live threads exist without their project rows, is wired from the root startup path, and has dedicated regression coverage; the shared project-create recovery also falls back to repaired snapshots. That matches the root cause already documented in this issue thread. Unless the empty-sidebar-after-restart symptom still reproduces on a current profile, this looks safe to close.

[...]

PR #44, opened 2026-04-20
Fix Windows desktop chrome and project recovery

justelson:
## Summary

This PR fixes three desktop issues:

1. [#35](https://github.com/Emanuele-web04/dpcode/issues/35): existing threads could disappear because desktop startup accepted a broken snapshot where live threads existed without matching live project rows.
2. [#43](https://github.com/Emanuele-web04/dpcode/issues/43): `project.create` could succeed without surfacing the created project, and retrying the same folder then failed with the duplicate workspace-root invariant.
3. Windows desktop chrome: the app was always taking the macOS title-bar path, which left the DP header offset incorrectly on Windows and Linux.

### Issue #35: thread/sidebar recovery on desktop startup

- Detect broken desktop snapshots where live threads exist without matching live project rows.
- Repair that snapshot shape during desktop bootstrap instead of only repairing fully empty snapshots.
- Prevent the sidebar from accepting that broken state and coming up empty even though thread data still exists.

### Issue #43: project-add recovery after silent success

- Centralize project recovery helpers so add-project and duplicate-create flows share the same read-model lookup behavior.
- Retry the read model briefly after project creation.
- Fall back to `repairState()` when polling still cannot surface the created project.
- Recover duplicate-create flows by reopening the existing project instead of leaving the UI empty and surfacing the duplicate workspace-root error on retry.

### Windows desktop chrome

- Split desktop window chrome by platform instead of always using the macOS inset title bar.
- Keep the native macOS traffic-light/title-bar behavior on macOS.
- Use a hidden title bar on Windows and Linux and expose renderer-driven window controls through the desktop bridge.
- Sync maximize/fullscreen state back to the renderer so maximize/restore icons stay in the correct state.
- Remove the extra right-edge gap in the custom desktop header so the control strip sits flush against the window edge.
- Render the same Windows controls across:
  - active chat header
  - empty-thread header
  - settings header
  - dedicated workspace header
- Update the custom control icons to the intended Windows-style minimize / maximize / restore / close states.

## Why

There were two separate recovery failures in the desktop app:

- In [#35](https://github.com/Emanuele-web04/dpcode/issues/35), desktop startup could accept a broken local/read-model snapshot and show an empty sidebar even though live threads still existed.
- In [#43](https://github.com/Emanuele-web04/dpcode/issues/43), the first add-project attempt could succeed internally but fail to appear in the UI, and retrying the same folder then produced:

`Orchestration command invariant failed (project.create): Project '<id>' already uses workspace root '<path>'`

Separately, the Windows title bar/header was still following the macOS chrome path, which left the branding and window controls misaligned on Windows/Linux.

## Validation

Ran targeted tests:

- `bun run test src/windowChrome.test.ts` in `apps/desktop`
- `bun run test src/lib/desktopWindow.test.ts src/lib/projectCreateRecovery.test.ts src/lib/desktopProjectRecovery.test.ts` in `apps/web`

Results:

- desktop: 1 file, 4 tests passed
- web: 3 files, 17 tests passed

## Notes

- I intentionally left unrelated/generated changes out of this PR, including `bun.lock`, `apps/web/src/routeTree.gen.ts`, and `apps/web/public/mockServiceWorker.js`.
- Before/after screenshots for the Windows chrome surfaces still need to be attached because part of this PR is a UI fix.

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -18,12 +18,13 @@ import {
   shell,
   systemPreferences,
 } from "electron";
-import type { MenuItemConstructorOptions } from "electron";
+import type { IpcMainInvokeEvent, MenuItemConstructorOptions } from "electron";
 import * as Effect from "effect/Effect";
 import type {
   DesktopTheme,
   DesktopUpdateActionResult,
   DesktopUpdateState,
+  DesktopWindowState,
 } from "@t3tools/contracts";
 import { autoUpdater } from "electron-updater";
 
@@ -61,6 +62,7 @@ import { isArm64HostRunningIntelBuild, resolveDesktopRuntimeInfo } from "./runti
 import { DesktopBrowserManager } from "./browserManager";
 import { registerBrowserIpcHandlers, sendBrowserState } from "./browserIpc";
 import { BrowserUsePipeServer } from "./browserUsePipeServer";
+import { resolveDesktopWindowChrome, resolveDesktopWindowState } from "./windowChrome";
 
 syncShellEnvironment();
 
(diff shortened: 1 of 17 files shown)


justelson:
Follow-up pushed in `5c87c016`.

Addressed the two Codex review findings:

- Made the workspace title explicitly non-draggable inside the custom Electron drag region, so double-click rename still works on Windows/Linux.
- Gated duplicate-project recovery behind `isDuplicateProjectCreateError`, so unrelated `project.create` failures no longer trigger recovery polling or `repairState()`.

Re-ran the targeted tests:

- `bun run test src/windowChrome.test.ts`
- `bun run test src/lib/projectCreateRecovery.test.ts src/lib/desktopWindow.test.ts src/lib/desktopProjectRecovery.test.ts`

justelson:
Follow-up pushed in `d3cf500c`.

Addressed the latest Codex review finding by applying the same macOS hidden-titlebar inset logic to `WorkspaceView` that the chat/settings headers already use.

This fixes the `isElectron && !supportsCustomDesktopTitleBar() && settings.sidebarSide === "right"` case, so the workspace header controls do not sit under the native macOS traffic-light buttons.

Re-ran the targeted tests:

- `bun run test src/windowChrome.test.ts`
- `bun run test src/lib/desktopWindow.test.ts src/lib/projectCreateRecovery.test.ts src/lib/desktopProjectRecovery.test.ts`

PR #58, opened 2026-04-24
Fix sidebar project sort reset

SunkenInTime:
## Summary
- remove the runtime legacy migration that rewrote project sort order from updated_at to manual on hook mount
- preserve explicit sidebar project sort choices when creating a new thread

## Testing
- Not run: focused test command failed locally because turbo is not installed/on PATH in this workspace.


SunkenInTime:
Comparison Footage:
Before: 
Recording 2026-04-24 111532
After:
Recording 2026-04-24 111347

PR #116, opened 2026-05-14
Fix select-all copy in the diff viewer

### LINKED ISSUE #60: [Bug]: Select all and copy from the diff file viewer, doesn't copy all the content into the clipboard
### Before submitting

### Area

apps/desktop

### Steps to reproduce

Create big enough file, in my case over 6k lines, try to open it in diff viewer and select all nd copy

### Expected behavior

all 6k lines are copied

### Actual behavior

around 150 lines are copied

### Impact

Blocks work completely

### Version or commit

_No response_

### Environment

_No response_

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

Zeus-Deus:
Closes #60

## Problem
The diff viewer is virtualized, so only a handful of rows are ever in the DOM. Selecting all and copying a large file (e.g. 6k lines) only put those few rendered rows on the clipboard.

## Fix
- Serialize the full diff from the parsed model instead of relying on mounted DOM rows.
- Intercept the copy event at the document level. The diff renders into shadow DOM, so the copy event never reaches the viewport element — Cmd/Ctrl+A inside the viewport arms the gesture, and the next copy writes the complete diff.
- The existing "Copy" button now uses the same serialized text.

## Testing
- Unit tests for the serializer (including a 6k-line diff) and the select-all gesture logic.
- Verified in-browser with a 12k-line working-tree diff: select-all + copy, the Copy button, and partial selections all behave correctly.

### DIFF
--- a/apps/web/src/components/DiffPanel.logic.test.ts
+++ b/apps/web/src/components/DiffPanel.logic.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from "vitest";
 
 import type { DraftThreadState } from "../composerDraftStore";
 import type { Thread } from "../types";
-import { resolveDiffPanelThread } from "./DiffPanel.logic";
+import { resolveDiffPanelThread, resolveDiffSelectAllArmed } from "./DiffPanel.logic";
 
 const PROJECT_ID = ProjectId.makeUnsafe("project-1");
 const THREAD_ID = ThreadId.makeUnsafe("thread-1");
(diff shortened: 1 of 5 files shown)


Marve10s:
Thanks for taking this on. The underlying bug looks valid: a virtualized diff cannot rely on the mounted DOM rows for full-file select-all copy.

Please tighten this PR before merge:

1. Clean up the added explanatory comments. Keep only comments that document non-obvious behavior that future maintainers could not infer from the code.
2. Keep the PR scoped to the select-all/copy bug. Do not add unrelated edits, formatting churn, or extra test expansion beyond what is needed for this bug.
3. Make sure the final fix is the smallest proper fix for the reported behavior: full-diff copy should use the parsed diff model, partial/native selections should continue to behave normally, and the Copy button should share the same serialization path.

Zeus-Deus:
Addressed the comment cleanup in a follow-up: #120. Kept it scoped to just that — one short line per function to match the rest of the file.

PR #117, opened 2026-05-14
Fix model picker breaking when Cursor CLI is unavailable

### LINKED ISSUE #103: [Bug]: Retry models, When cursor CLI is not available.
### Before submitting

### Area

apps/web

### Steps to reproduce

https://github.com/user-attachments/assets/4fcb1b42-8037-4b4d-8f9e-4e41d35104ce

If Curser CLI is not installed or is not authenticated.

Error: Retry models

### Expected behavior

Disable cursor and show other models selector.

### Actual behavior

Disabling whole model selector and Error: retry models.

### Impact

Major degradation or frequent failure

### Version or commit

_No response_

### Environment

_No response_

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

Zeus-Deus:
Fixes #103

When the Cursor CLI isn't installed or authenticated, Cursor's model discovery fails — and that failure was rejecting the whole models query, leaving the model picker stuck in a loading/error state for every provider.

### Changes

- `providerModelsQueryOptions` now catches discovery failures and resolves to an empty result instead of rejecting. A failing CLI degrades only that provider; every other provider keeps its models.
- The Cursor/Kilo discovery skeleton is gated on `isLoading` instead of `isFetching`, so background refetches don't re-blank the picker once discovery has settled.
- Added tests covering a single provider failing without affecting the others, plus a missing `kilo` key in an existing test fixture.

### Testing

- `bun lint` and `bun run test` pass.
- Verified in the app: with Cursor's binary path pointed at a missing binary, the picker opens normally, Cursor falls back to its static model list, and every other provider works as usual.

Note: `bun typecheck` still reports one pre-existing, unrelated error in `composerDraftStore.ts` (Pi provider), and two server tests fail for environment reasons — both already present on `main` and untouched here.

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -1477,10 +1477,14 @@ export default function ChatView({
   const hasResolvedCursorModelDiscovery =
     cursorDynamicModelsQuery.data?.source === "cursor.cli" &&
     (cursorDynamicModelsQuery.data.models.length ?? 0) > 0;
+  // Only the very first discovery attempt (no settled result yet) should gate
+  // the picker. Once discovery resolves — including a fault-isolated "error"
+  // result when the Cursor CLI is unavailable — background refetches must not
+  // re-blank the shared model picker.
   const cursorModelDiscoveryPending =
     cursorModelDiscoveryEnabled &&
     !hasResolvedCursorModelDiscovery &&
-    (cursorDynamicModelsQuery.isLoading || cursorDynamicModelsQuery.isFetching);
+    cursorDynamicModelsQuery.isLoading;
   const kiloModelDiscoveryEnabled =
     selectedProvider === "kilo" || lockedProvider === "kilo" || isModelPickerOpen;
   const hasResolvedKiloModelDiscovery =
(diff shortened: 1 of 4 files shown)


Marve10s:
Thanks for working on this. I reviewed the bug path and the issue is real: when Cursor model discovery rejects because the CLI is missing/unauthenticated/unavailable, the shared model picker can end up in the discovery error path instead of degrading Cursor only.

Please tighten this PR before merge:

1. Clean up the added explanatory comments. Keep only comments that are necessary for non-obvious behavior.
2. Keep the PR scoped to the model-discovery failure bug. Do not add unrelated edits, formatting churn, or extra test expansion beyond what is needed for this bug.
3. Apply the proper fix for the loading-state issue in the current patch. `providerModelsQueryOptions` uses `placeholderData`, and with TanStack Query the first unresolved fetch can report `isLoading: false`, `isFetching: true`, and `isPlaceholderData: true`. Changing Cursor/Kilo discovery pending from `isLoading || isFetching` to only `isLoading` means initial discovery may no longer show the loading state and can expose fallback static models while discovery is still running. Please distinguish the initial placeholder fetch from later background refetches, and add/adjust focused coverage for that behavior.

Zeus-Deus:
Addressed all 3 points in #121. Loading state now uses isFetching && isPlaceholderData so the initial fetch is caught but background refetches aren't, plus a test for it.

PR #132, opened 2026-05-28
Add Claude Opus 4.8 ultracode support

Marve10s:
Summary
- Add Claude Opus 4.8 with 200k and 1M context options.
- Include the full supported Claude effort menu, including ultracode.
- Update the Claude Agent SDK to a version that exposes the ultracode session setting.
- Map ultracode to xhigh effort plus the Claude Code ultracode setting, which enables dynamic workflow orchestration.

Validation
- Not run per request.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -23,7 +23,8 @@
     "test": "vitest run"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


Marve10s:
Closing in favor of #133, which is the narrow hotfix against main with only the SDK/package updates.

PR #138, opened 2026-06-02
fix(desktop): hydrate Windows env from the registry

ramarivera:
## What Changed

`syncShellEnvironment` now handles Windows. It previously early-returned on any
non-macOS/Linux platform, so the desktop never hydrated its environment there.
Added a `win32` branch that reads the persisted **User + Machine** registry
environment and:

- merges it into `PATH`
- fills in any variable not already set in the process env (e.g. `CLAUDE_CONFIG_DIR`)

The registry read sits behind an injectable reader (`readWindowsPersistentEnvironment`),
so the merge logic is unit-tested without touching a real registry.

## Why

On Windows, GUI apps inherit Electron's launch-time environment block, not the
user's current shell/registry environment. So a CLI provider the app spawns can
miss config the user's terminal already has — most visibly `CLAUDE_CONFIG_DIR`,
which Claude Code needs to locate its credentials. The symptom was Claude showing
a false "not authenticated" / greyed-out **Sign In** even though `claude` was
logged in from a terminal on the same machine. macOS and Linux already avoid this
via the login-shell sync; this brings Windows to parity.

## Tests

- `packages/shared`: unit tests for `mergeWindowsScopes` (scope precedence + PATH
  composition) and `readWindowsPersistentEnvironment` (registry-dump parsing and
  bad-JSON fallback).
- `apps/desktop`: `syncShellEnvironment` win32 tests (hydrates PATH + missing vars;
  preserves variables already in the environment).
- Full `shared` + `desktop` suites, `typecheck`, `oxlint`, and `oxfmt --check` pass.

The injectable seams and JSON parsing are covered cross-platform; the actual
PowerShell registry read only runs on Windows, so that live path hasn't been
exercised on CI hardware.

## Checklist

- [ ] No UI changes
- [ ] No animation/interaction changes

🤖 *This content was generated with AI assistance using Claude Opus 4.8.*

### DIFF
--- a/apps/desktop/src/syncShellEnvironment.test.ts
+++ b/apps/desktop/src/syncShellEnvironment.test.ts
@@ -174,7 +174,47 @@ describe("syncShellEnvironment", () => {
     expect(env.PATH).toBe("/opt/homebrew/bin:/usr/bin");
   });
 
-  it("does nothing outside macOS and Linux", () => {
+  it("hydrates PATH and missing variables from the Windows registry", () => {
+    const env: NodeJS.ProcessEnv = {
+      PATH: "C:\\Windows\\system32",
+    };
+    const readWindowsEnvironment = vi.fn(() => ({
+      PATH: "C:\\Windows\\system32;C:\\Users\\ramar\\.local\\bin",
+      CLAUDE_CONFIG_DIR: "C:\\Users\\ramar\\.config\\claude",
+    }));
+
+    syncShellEnvironment(env, {
+      platform: "win32",
+      readWindowsEnvironment,
+    });
+
+    expect(readWindowsEnvironment).toHaveBeenCalledTimes(1);
+    expect(env.PATH).toBe("C:\\Windows\\system32;C:\\Users\\ramar\\.local\\bin");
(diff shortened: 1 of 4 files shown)


ramarivera:
Tested the `win32` branch on a real Windows 11 machine — CI can't exercise it since it shells out to `powershell.exe`, so I ran it on hardware.

Ran this branch's `readWindowsPersistentEnvironment` through bun 1.3.14:

- merged Machine + User registry scopes → 57 vars
- PATH composed and merged correctly (length 2867)
- non-PATH vars hydrated, e.g. `CLAUDE_CONFIG_DIR` present

The full `powershell dump → JSON.parse → mergeWindowsScopes` chain works on Windows. Separately confirmed the payoff: a process whose env carries that hydrated `CLAUDE_CONFIG_DIR` gets `claude auth status` → `loggedIn: true`, where the desktop previously showed a false greyed-out **Sign In**.

I didn't rebuild the full Electron app on Windows — the `syncShellEnvironment` → reader → `process.env` wiring is covered by the unit tests in this PR; the part that can only run on Windows (the registry read) is what I verified on the box.

🤖 *This content was generated with AI assistance using Claude Opus 4.8.*

PR #146, opened 2026-06-03
Add linked provider continuation handoff

jgarbarino3:
## Summary

Adds a linked continuation flow for conversations that are already bound to a provider.

The provider/model picker still lets you change models within the current provider, and now also offers a simple "Continue with..." path for starting a linked thread with another provider. The new thread receives a compact context packet and the original provider session stays unchanged.

## What changed

- Add "Continue with..." options to the locked provider/model picker.
- Add a confirmation dialog before creating the linked continuation thread.
- Copy safe draft state into the new thread: current prompt, assistant selections, terminal contexts, and current image attachments.
- Keep queued follow-up turns on the original thread.
- Add lightweight source/target continuation links in the header and sidebar.
- Add a collapsed context preview so the user can inspect what will be handed off.

## Verification

- Focused web tests passed.
- Shared handoff context tests passed.
- Server handoff/runtime tests passed.
- Provider picker and handoff dialog browser tests passed in Chromium.
- `bun run typecheck`, `bun run fmt:check`, `bun run lint`, and `git diff --check` passed.

## Follow-up

A later iteration could explore switching providers inside the same visible thread with explicit provider/session segments. This PR intentionally keeps the safer linked-thread model.


Emanuele-web04:
I'm sorry I don't understand the difference from the Handoff feature I already have

jgarbarino3:
> I'm sorry I don't understand the difference from the Handoff feature I already have

Yeah..my bad. I was trying to allow for provider switching within the same conversation. So this was more of a work in progress and I didn't mean to make it a PR yet.

PR #147, opened 2026-06-03
[codex] Fix settings back navigation and legacy sidechat migration

Blueemi:
## What changed

- Added a reconciliation migration that reruns the sidechat source column migration for legacy databases whose migration tracker reused ID 33.
- Centralized settings/back-to-thread target resolution and reused it for both settings back navigation and the sidebar thread/workspace switch.
- Added focused tests for the legacy migration repair and settings back target selection.

## Why

Imported legacy databases can appear fully migrated while still missing `projection_threads.sidechat_source_thread_id`, which breaks read-model queries that expect the column. Settings back navigation also needed to avoid stale thread/split-view targets and fall back predictably.

## Validation

- `bun run test src/components/Sidebar.logic.test.ts` from `apps/web`: 71 passed
- `bun run test src/persistence/Migrations/038_ReconcileLegacySidechatSource.test.ts` from `apps/server`: 2 passed

Note: full `bun fmt`, `bun lint`, and `bun typecheck` were not run because repo instructions say not to run those heavyweight checks unless explicitly requested in the current conversation.


Emanuele-web04:
Superseded by #148, which keeps these fixes and adds the same-class pinned-thread migration repair.

PR #162, opened 2026-06-09
Split composer model and options pickers

Marve10s:
## Summary
- Split the empty-thread composer control into a dedicated model picker and a separate options picker.
- Keep started and existing threads on the combined model/options trigger.
- Keep effort, context, and speed controls in the options picker for empty threads.
- Update picker browser expectations for the new labels.

## Verification
- git diff --check

Focused browser tests were not run in this follow-up.


Emanuele-web04:
test for poke

PR #165, opened 2026-06-10
Improve editor view, composer layout, and terminal env

Marve10s:


Emanuele-web04 (commented):
Reviewed the full diff (5 parallel review passes over editor view, diff panel, chat/routes, composer, and server/contracts), then pushed two commits: a merge of `main` and a fix commit. Summary:

**Why CI was red:** `Release Smoke` failed because the branch predated the `@pierre/diffs` 1.2.8 pin on main (`@pierre/theming@0.0.1` 404s on npm). Merging main in fixes it — not a fault of this PR.

**Fixed in `14f7db8db`:**
- **Symlink escape in the new workspace read/list RPCs** (`workspaceFileSystem.readFile`, `listWorkspaceDirectories`): containment was string-only (`path.resolve`/`path.relative`), so a symlink inside the workspace pointing outside passed the check and was followed. Added realpath-based containment (in-root symlinks still work) and switched `readFile` to stat through the open handle to remove the stat→open race. Tests added.
- **Hook-order crash in the syntax-highlight components**: `SyntaxHighlightedFileContents` (and the pre-existing `SuspenseShikiCodeBlock` it was modeled on) called `useMemo`/`useEffect` after the cache-hit early return — once the cache filled, the next re-render threw "Rendered fewer hooks than expected", and the unkeyed error boundary then latched every file into the plain fallback. Split the uncached path into its own component and keyed the boundary by file path.
- **TraitsPicker trigger regressions**: `resolveTraitsTriggerSummary` dropped the `defaultContextWindow` guard (the context-window label showed permanently for any provider with one) and the fast-only `Fast`/`Default` primary label. Restored the original semantics.
- Smaller items: backtick-fence escaping in `buildDiffSelectionReference`; reuse `basenameOfPath` instead of a local duplicate; remount `DiffPanelToolbar` per thread so "Show more" pagination doesn't leak across threads; renamed the contracts constant shared by read+write path schemas.

**Noted but left as-is** (pre-existing or judgment calls): the localStorage re-read effect in `EditorRailTabs` runs on every title/provider change (harmless but churns; consider guarding on `projectId`), the `EditorDiffOptionsMenu` trigger has redundant manual open handlers, and the file-preview context menu silently no-ops outside the desktop app despite the comment promising a DOM fallback.

Verified locally: `bun fmt`, `bun lint` (0 errors), `bun typecheck`, `bun run test` (1187 passed).

PR #172, opened 2026-06-11
Fix editor mode production feedback

Marve10s:
## Summary

Fixes the production feedback from the new Editor mode rollout:

- render supported local images in the Editor file viewer and diff sidebar instead of trying to read them as text
- add a Markdown preview/source toggle for Markdown files
- reserve Windows caption-control space in the Editor top bar so native window buttons do not overlap app controls
- add a compact current-project switcher beside the project title/path, with recent projects only and tuned hover/active states

## Impact

Editor mode now handles image and Markdown files more predictably, the top bar stays usable on Windows, and switching between recent projects is available without leaving the Editor surface.

## Validation

- `bun run test src/components/EditorWorkspaceView.test.tsx src/lib/diffRendering.test.ts src/hooks/useDesktopTopBarGutter.test.ts`
- `git diff --check`


Marve10s:
Closing in favor of #173, which is rebased onto the current main branch and contains only the Editor-mode production feedback fixes.

PR #225, opened 2026-06-23
[codex] fix automation audit findings

Emanuele-web04:
## Summary

- cleans up newly created standalone automation worktrees when cancellation or `thread.create` failure happens before durable thread ownership exists
- guards live automation definition/run upserts by `updatedAt` so stale stream events cannot roll back newer cache rows
- adds a focused pending completion-evaluation migration/index and consolidates duplicated scheduler-critical SQL through a shared view
- bounds the completion-evaluation queue while preserving DB reconciliation for pending rows

## Validation

- `cd apps/server && bun run test src/automation/Layers/AutomationService.test.ts src/persistence/Layers/AutomationRepository.test.ts src/persistence/Migrations.test.ts src/persistence/Migrations/044_Automations.test.ts`
- `cd apps/web && bun run test src/routes/-automations.shared.test.tsx`
- `git diff --check`

## Notes

I did not run `bun fmt`, `bun lint`, or `bun typecheck` because this thread explicitly limited verification to focused tests unless those heavyweight checks are requested.

PR #228, opened 2026-06-23
refactor(shared): tighten deriveAssociatedWorktreeMetadata param types to match its undefined handling

tszaks:
Widens the three `associated*` params of `deriveAssociatedWorktreeMetadata` (and its sibling `deriveAssociatedWorktreeMetadataPatch`) from `string | null` to `string | null | undefined`, matching the `x !== undefined` branching the functions already perform and the `Schema.optional` thread fields they consume under `exactOptionalPropertyTypes`. Behavior-preserving (a call-site `?? null` would have erased the worktreePath-derivation path).

Context: I branched this before your #225/#226 merges, where it fixed a `tsc` failure at `ChatView.tsx:1423`. Your attachment merge has since reworked that call site, so this is now a standalone type-hygiene improvement rather than a compile fix.

Heads-up unrelated to this PR: on current `main`, `bun typecheck` fails in `apps/server/src/automation/Layers/AutomationService.test.ts` lines 603 and 665 (`createdWorktrees[0]?.newBranch` is `string | undefined` where a `string` is expected) from the cleanupUnattachedWorktree test in #225 - flagging in case CI did not catch it.

### DIFF
--- a/packages/shared/src/threadWorkspace.ts
+++ b/packages/shared/src/threadWorkspace.ts
@@ -114,9 +114,13 @@ export function isScratchWorkspacePath(filePath: string): boolean {
 export function deriveAssociatedWorktreeMetadata(input: {
   branch?: string | null;
   worktreePath?: string | null;
-  associatedWorktreePath?: string | null;
-  associatedWorktreeBranch?: string | null;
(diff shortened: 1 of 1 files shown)


tszaks:
@codex review

PR #229, opened 2026-06-23
fix(automations): re-check turn ownership before resuming an approval wait

tszaks:
**Race:** entering `waiting-for-approval` already requires `runTurnOwnsPendingInput`, but the EXIT back to `running` checked only `run.status` + thread/message/turn ids. Once a later, foreign turn takes over the same heartbeat thread and the original approval clears, reconcile flips the run back to `running` even though it no longer owns the turn, resurrecting a run that should stay parked.

**Fix:** add `runTurnOwnsPendingInput(run, shell, turn)` to the exit condition, mirroring the entry guard. Standalone runs short-circuit to `true`, so their resume path is unchanged (covered by the existing "cleared approval wait back into running" test).

**Test:** a heartbeat run whose thread's latest turn becomes an unrelated turn stays `waiting-for-approval` instead of resuming. Complements the existing completion-path ownership test. 77/77 pass.

### DIFF
--- a/apps/server/src/automation/Layers/AutomationService.test.ts
+++ b/apps/server/src/automation/Layers/AutomationService.test.ts
@@ -970,6 +970,72 @@ layer("AutomationService", (it) => {
     }),
   );
 
+  it.effect("does not resume a waiting-for-approval run from an unrelated newer turn", () =>
+    Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


tszaks:
@codex review

PR #230, opened 2026-06-23
fix(automations): don't let stop-evaluation clobber a run's archive/read state

tszaks:
**Race (most data-corrupting of the lifecycle bugs):** `recordCompletionEvaluation` reloads the run, then writes the whole `result_json` blob via `markRunResult`. `automationCompletionRunResult` spreads `...baseResult`, so `archivedAt`/`unread` come from the *reloaded* run. A user archive or mark-read landing between the reload and the write is silently reverted - a background heartbeat can un-archive a run the user just filed away.

**Fix:** add `markRunCompletionResult` to the repository. It writes the new result but `json_set()`s `$.archivedAt` and `$.unread` from the **current row**, closing the window atomically SQL-side. `unread` round-trips through `json()` so it stays a JSON boolean (not the 0/1 `json_extract` yields); a legacy/null `unread` falls back to the incoming value rather than defaulting to unread. `recordCompletionEvaluation` now calls it instead of `markRunResult`.

**Tests:** a completion eval landing after an archive keeps `archivedAt`/`unread`; a read-but-not-archived run keeps `unread=false` with `archivedAt` staying null. 24/24 repository + 76/76 service tests pass.

### DIFF
--- a/apps/server/src/automation/Layers/AutomationService.ts
+++ b/apps/server/src/automation/Layers/AutomationService.ts
@@ -1074,7 +1074,7 @@ export const AutomationServiceLive = Layer.effect(
         const latestRun = yield* latestRunForCompletionResult(input.run);
         const updatedAt = isoNow();
         const updated = yield* automationRepository
-          .markRunResult({
+          .markRunCompletionResult({
             id: latestRun.id,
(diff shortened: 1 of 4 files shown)


tszaks:
@codex

PR #233, opened 2026-06-23
fix(automations): always confirm LLM-generated automations before creating them

tszaks:
**Safety hole:** a high-confidence, thread-scoped *generated* (LLM-interpreted) intent with no completion policy resolves to `requiresReview=false`, so the draft gate (`needsDraftReview = resolution.requiresReview || ...`) lets it **auto-create with no confirmation dialog**. A misread chat message could silently spin up a recurring background automation.

**Fix:** force `requiresReview = true` for `source === "generated"` in `resolveChatAutomationIntent`. Because `needsDraftReview` is `requiresReview || ...`, the resolver is the single choke point and this can only ever *add* a confirmation, never remove one.

**Preserves your deterministic gating:** explicit, locally-parsed intents keep their finer-grained logic, including the intentional bounded-fast-loop auto-submit (`isAutoSubmittableBoundedFastLoop`) — that path skips generation entirely, so it never reaches the generated branch.

**Judgment call for you:** this also makes an LLM-*generated* bounded fast loop require review (only the *deterministic* one auto-submits). That seemed right since generated intents are the model's interpretation, but happy to scope it differently if you prefer.

Tests: strengthened the existing generated-intent test to assert `requiresReview`, added an explicit invariant test; composerAutomation + automationDraft suites still green.

### DIFF
--- a/apps/web/src/lib/automationIntent.test.ts
+++ b/apps/web/src/lib/automationIntent.test.ts
@@ -508,6 +508,9 @@ describe("parseChatAutomationIntent", () => {
     expect(resolved).toMatchObject({
       source: "deterministic",
       mode: "heartbeat",
+      // The schedule parsed deterministically, but name/prompt are LLM-rewritten, so the
+      // draft must still go through human review even though needsConfirmation was false.
+      requiresReview: true,
       intent: {
         name: "Generated",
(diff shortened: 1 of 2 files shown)


tszaks:
@codex review

PR #234, opened 2026-06-23
chore(automations): make recovery failures and lease contention observable

tszaks:
`reconcileActiveRuns` and `recoverPendingRuns` swallowed every failure with `Effect.catch(() => Effect.void)`, so a run that fails to reconcile or recover vanished silently. The scheduler-lease loser also returned `[]` with no trace, hiding multi-instance contention.

**Change (log-only, behavior unchanged):**
- `logWarning` (with automationId/runId) in the three recovery catches.
- `logDebug` when the scheduler lease is not acquired.
- New `recoveryErrorMessage` helper unwraps `AutomationServiceError.cause` (the static wrapper message alone is useless) on top of your existing redacting `errorMessage`.

Scoped to the recovery/lease paths - deliberately does **not** touch the queue/worker/scheduled-run logging you already added elsewhere. The catches still recover to void; this only makes the failures visible. 76/76 pass.

### DIFF
--- a/apps/server/src/automation/Layers/AutomationService.ts
+++ b/apps/server/src/automation/Layers/AutomationService.ts
@@ -111,6 +111,15 @@ function errorMessage(cause: unknown): string {
   return redactSecrets(raw).slice(0, AUTOMATION_ERROR_MAX_CHARS);
 }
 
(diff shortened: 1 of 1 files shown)


tszaks:
@codex review

PR #235, opened 2026-06-23
[codex] Fix profile skill usage counts

Emanuele-web04:
## Summary

Fixes profile skill usage counting so the Profile page reflects real `$skill` and `/skill` usage.

## Root Cause

Profile stats excluded retention-hidden threads by requiring `t.deleted_at IS NULL`, even though retention intentionally keeps those rows for history. Skill aggregation also collapsed repeated `$skill` / `/skill` tokens in the same user message to one run.

## Changes

- Include retention-hidden threads in profile stats while still excluding manually deleted threads and deleted projects.
- Count repeated slash/dollar skill tokens inside a single prompt without double-counting structured skill references.
- Add focused server tests for retention-hidden threads and repeated skill invocations.

## Validation

- `bun run --cwd apps/server test src/profileStats.test.ts`
- `git diff --check`

### DIFF
--- a/apps/server/src/profileStats.test.ts
+++ b/apps/server/src/profileStats.test.ts
@@ -66,6 +66,30 @@ describe("ProfileStatsQuery", () => {
     ]);
   });
 
+  it("counts repeated slash and dollar skill tokens without double-counting structured echoes", () => {
+    expect(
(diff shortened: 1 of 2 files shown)


Emanuele-web04:
@codex review

PR #247, opened 2026-06-24
fix(typecheck): restore green typecheck on main

tszaks:
`bun typecheck` currently fails on `main` with 9 errors across `shared`, `server`, and `web`. All are type-level fixes; no runtime behavior changes.

- **shared/windowsProcess.ts** — `result.stdout` was typed `any` (via the injectable `spawnSync`), so the `.map`/`.find` callbacks were implicit-any; coerce it to a string first. Also widen `WindowsSafeProcessInput`'s optional fields to `T | undefined` so callers can pass `{ cwd, env }` under `exactOptionalPropertyTypes` (fixes the `processRunner` / `AcpSessionRuntime` / `opencodeRuntime` call sites).
- **server/ProviderService.ts** — annotate `stopRuntimeSessionInternal`'s `rawInput` (was implicit any) with the parameter type of `stopRuntimeSession`.
- **server/ProviderService.test.ts** — assert the captured `releaseListSessions` is present before invoking (the existing `typeof === "function"` check already guarantees it at runtime).
- **web/toolCallDetails.ts** — cast the narrowed `object` to `Record<string, unknown>` in `asRecord`.

Verified: `bun fmt`, `bun lint` (0 errors), `bun typecheck` (8/8 green), and the ProviderService suite (37/37) all pass.

### DIFF
--- a/apps/server/src/provider/Layers/ProviderService.test.ts
+++ b/apps/server/src/provider/Layers/ProviderService.test.ts
@@ -168,8 +168,8 @@ function makeFakeCodexAdapter(provider: ProviderKind = "codex") {
       Effect.succeed({ threadId, turns: [] }),
   );
 
-  const compactThread = vi.fn((_threadId: ThreadId): Effect.Effect<void, ProviderAdapterError> =>
-    Effect.void,
+  const compactThread = vi.fn(
+    (_threadId: ThreadId): Effect.Effect<void, ProviderAdapterError> => Effect.void,
   );
 
   const stopAll = vi.fn(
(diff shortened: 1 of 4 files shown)


tszaks:
@codex review

tszaks:
Heads up @Emanuele-web04: this is a small, type-only fix that gets `bun typecheck` green again on `main` (it's currently red in 9 spots across shared/server/web from a few recent commits). Codex reviewed it clean. No rush at all, but merging it unblocks a couple of follow-ups I'm lining up. Thanks!

PR #250, opened 2026-06-24
refactor: tidy PR #247 provider type cleanup

Emanuele-web04:
Follows up #247.

## Summary
- Names the provider runtime-stop function, input, and return types once instead of repeating the long inline contract type.
- Replaces repeated deferred listSessions callback type spellings and non-null assertions in ProviderService race tests with a small assertion helper.

## Verification
- bun run --cwd apps/server test src/provider/Layers/ProviderService.test.ts
- bun run typecheck
- git diff --check

## Review
- Same-directory /review thread 019efc09-9406-7903-800b-66bdb1f6c1df: no actionable findings.
- Second same-directory /review thread 019efc0b-1cd1-7fe1-b2d8-7b6db17819d9: no actionable findings.

### DIFF
--- a/apps/server/src/provider/Layers/ProviderService.test.ts
+++ b/apps/server/src/provider/Layers/ProviderService.test.ts
@@ -62,6 +62,18 @@ type LegacyProviderRuntimeEvent = {
   readonly [key: string]: unknown;
 };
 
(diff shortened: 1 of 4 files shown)


Emanuele-web04:
@codex review

PR #243, opened 2026-06-24
fix(automations): harden approval fallback coverage

Emanuele-web04:
Follow-up to #242 (`feat(automations): one-click approval for blocked automations`).

This branch starts from the current PR #242 head and keeps the follow-up isolated on `codex/pr-242-approval-gate-fixes`.

What changed:
- Refactors the existing-automation approval helper to avoid the warning-id-to-risk-id cast and reuse the shared warning-to-risk mapper for the full persisted approval payload.
- Adds focused coverage for standalone `auto` worktree fallback so one-click approval persists `local-checkout` as expected.

Verification:
- `bun run --cwd apps/web test automationDraft.test.ts`
- `git diff --check`

Review passes:
- Same-directory check-code review pass 1: no actionable findings.
- Same-directory check-code review pass 2: no actionable findings.

PR #251, opened 2026-06-24
fix(automations): restore approval fallback

Emanuele-web04:
Follows up #243 after #245 superseded the server-side run gate work. #245 is now the source of truth for server enforcement; this PR restores the missing web fallback for already-created automations that still need one-time approval.

What changed:
- Adds `automationApprovalGaps` for existing automation definitions so the detail panel can surface missing approvals and persist the full required risk set.
- Adds an approval banner with Approve and Approve & run now actions, while keeping Run now disabled until approval saves.
- Patches `maxIterations` when approving legacy fast-interval automations, including the edge where `fast-interval` is already acknowledged but the required safety cap is missing.
- Keeps the final diff web-only against `main`; superseded #243 server edits were neutralized in favor of merged #245.

Verification:
- `bun run --cwd apps/web test automationDraft.test.ts`
- `git diff --check`
- Same-directory `/review` pass 1: clean
- Same-directory `/review` pass 2: clean

Did not run `bun fmt`, `bun lint`, or `bun typecheck` per this thread/repo instruction.


Emanuele-web04:
@codex review

PR #256, opened 2026-06-26
fix: persist review walkthrough outputs

Tbsheff:
# Description

This fixes the review walkthrough and review board state regressions reported in the review surface:

- Persist generated walkthrough outputs through the SQLite review cache, keyed by canonical PR number, token identity, patch signature, and generation settings so navigating away and back does not force a full regeneration.
- Preserve cached walkthrough rows across the token-identity migration and add a forward repair migration for databases that already recorded the earlier migration.
- Hydrate walkthrough React Query caches from domain events using the full query-result shape, including active generation-setting query keys.
- Disable nested sticky diff file headers inside review file diff blocks to avoid bad rendering while scrolling the review diff section.
- Keep review board/list data stable during lane load-more and facet changes; board-lane hydrate now returns list-style pagination metadata so load-more is deterministic.

Related issue: review walkthrough results were treated like transient UI output instead of durable review data, causing reruns after leaving and returning to a walkthrough. Review board hydration also had a hidden server/client contract mismatch around lane pagination.

Dependencies required: none.

## Test Plan

How to test:

- [x] `git diff --cached --check`
- [ ] `bun run test apps/web/src/lib/reviewReactQuery.test.ts apps/server/src/review/Layers/ReviewSource.test.ts apps/server/src/persistence/Migrations/049_ReviewWalkthroughCacheTokenIdentity.test.ts apps/web/src/components/review/PullRequestList.browser.tsx apps/web/src/components/review/ReviewBoard.browser.tsx` blocked locally because `turbo` is not installed in this worktree (`/bin/bash: turbo: command not found`).

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have walked through my test plan
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests to cover new or changed behavior

---

Compound Engineering
GPT-5


Tbsheff:
Closing because this PR was opened against the upstream repo by mistake; recreating it on the fork.

PR #264, opened 2026-06-27
feat: add /export slash command to download a thread as a ZIP archive

jld2700:
## What

Adds an app-level `/export` slash command that downloads the current thread as a portable, compressed ZIP archive — mirroring the `/export` affordance of agent CLIs.

## Why

There's no way to get a conversation out of Synara as a single, shareable artifact today. `/export` gives users a one-keystroke download of the current thread as a ZIP, without expanding product scope or touching provider internals. It's provider-agnostic and reads only Synara-owned data (the orchestration projection), so it works for every provider.

## What's in the archive

| File | Contents |
| --- | --- |
| `thread.json` | Structured thread projection — messages (role/text/source/timestamps), model, runtime mode, timestamps |
| `transcript.md` | Human-readable Markdown rendering of the conversation |

Filename: `synara-thread-<title-slug>-<YYYYMMDD>.zip`.

## How it's wired (minimal, one touch per layer)

- **`packages/shared`** — register `"export"` in `BUILT_IN_COMPOSER_SLASH_COMMANDS`.
- **`apps/web`** — add the `/export` definition and expose it for every provider (including `claudeAgent`, kept app-level like `/side` and `/automation`, since Synara owns the thread transcript rather than forwarding to Claude's native `/export`). The composer branch in `useComposerSlashCommands` intercepts it and triggers a browser download via the existing `downloadUrlAsBlob` helper against the new server route.
- **`apps/server`** — `GET /api/thread-export?threadId=…`, guarded exactly like the other binary GET routes (favicon/attachments). It reads the orchestration snapshot, builds the ZIP, and streams it back with `Content-Disposition: attachment`.

The ZIP writer is hand-rolled (local file headers + central directory + EOCD) using `node:zlib` (`deflateRawSync` + `crc32`) — **no new dependency added**. The compression method is chosen per-entry (store vs deflate).

## Verification

- `bun run typecheck` passes for `packages/shared`, `apps/server`, and `apps/web`.
- `oxlint` (with `--report-unused-disable-directives`) and `oxfmt --check` are clean on all changed files.
- New + updated vitest suites pass. The archive writer has a round-trip test that parses the produced ZIP (EOCD + central directory) and inflates each entry via `inflateRawSync`, proving the byte layout is valid without depending on a host `unzip` binary.
- The ZIP layout was additionally verified locally with `unzip -l` / `unzip -p`.

## Scope

No new dependencies, no new WebSocket contracts, no changes to any provider adapter. 5 files modified, 2 added. Happy to shrink or rework any of it.

### DIFF
--- a/apps/server/src/http.ts
+++ b/apps/server/src/http.ts
@@ -7,6 +7,7 @@ import {
   AuthCreatePairingCredentialInput,
   AuthRevokeClientSessionInput,
   AuthRevokePairingLinkInput,
+  ThreadId,
 } from "@t3tools/contracts";
 import { EDITOR_ICON_ROUTE_PATH } from "@t3tools/shared/editorIcons";
 import { DateTime, Effect, Exit, FileSystem, Layer, Path, Schema, Stream } from "effect";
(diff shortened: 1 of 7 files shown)


Emanuele-web04:
@codex review

Emanuele-web04:
Thanks for the contribution — the feature and the hand-rolled ZIP writer are solid. Review found a few issues (desktop auth via the legacy `?token=` param, markdown escaping corrupting code in transcripts, the server filename being unreachable, a defect on whitespace `threadId`, and sync deflate on the event loop), so this is superseded by #297, which carries your commit plus the fixes. Closing in favor of that PR.

PR #267, opened 2026-06-28
Chill hover states

ojowwalker77:
## What Changed

Chilled out a few hover states that felt too jumpy:

- lighter shared hover background
- active sidebar rows now stay stronger than hover rows
- slightly slower/easier hover color transitions
- sidebar hover cards keep the same frosted style, but no longer flash while opening
- project folder icon no longer swaps into a pin icon on hover
- transcript selection toolbar no longer grows on hover

## Why

Some of the hovers felt a bit too aggressive: too bright, too fast, or like things were glitching instead of transitioning.

This keeps the UI behavior the same, but makes hover feel calmer and less distracting. The hover-card fix is only for the flicker; it does not change the final frosted look.

I'll attach a short before/after video.

Refs:
- Material hover/state layers: https://m3.material.io/foundations/interaction/states/state-layers
- Apple HIG buttons/hover affordances: https://developer.apple.com/design/human-interface-guidelines/buttons

If preferred, I can follow up by moving some of this behind a Reduce Motion setting.

## Checklist

- [x] I included a video for animation/interaction changes


ojowwalker77:
https://github.com/user-attachments/assets/a5527552-a7c8-4499-8c38-f14923a9b19b

Emanuele-web04:
bro you removed the good part, the animations lol

ojowwalker77:
> bro you removed the good part, the animations lol

there is a reason for why accessibility is a thing on UI/UX lmao

ojowwalker77:
as I said this could be under a reduce motion setting, if you prefer

Issue #10, opened 2026-04-14
[Bug]: Loader persist despite response already fully streamed


jacerchetoui56:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

It does not happen always, but it happened to me the moment I ran two threads at the same time each having a different model (claude and codex) 

### Expected behavior

Spinner (in the sidebar) and "Working for xx" indicator should disappear right after the answer is provided.

### Actual behavior

the spinners keep on spinning and the "Working for xxx" is always running instead of disappearing. 

### Impact

Minor bug or occasional failure

### Version or commit

main@0.0.22

### Environment

macOS 26.5

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[65061.png](https://github.com/user-attachments/assets/42218b66-a16d-40ca-96e4-58eaff48d52a)
[32825.png](https://github.com/user-attachments/assets/26c9ea21-4f95-4e14-b70c-50d17a8d2681)

### Workaround

it's non blocking, so it's a minor issue.

jacerchetoui56:
I guess that was a problem with the modal taking too long doing something else and I thought it was done.

PR #268, opened 2026-06-28
Devin ACP provider

kartikkabadi:
## What Changed

Adds **Devin as an ACP provider** end-to-end, built on the official `@agentclientprotocol/sdk` through the shared `AcpSessionRuntime`, plus the P0/P1 hardening surfaced by dimensional review and live UI/UX testing.

Core runtime:
- **Devin ACP adapter** (`apps/server/src/provider/Layers/DevinAdapter.ts`) — turns, modes, plans, permissions, elicitations, attachments, and session resume/cleanup wired through `AcpSessionRuntime`.
- **Model compatibility** (`apps/server/src/persistence/modelSelectionCompatibility.ts`) — the provider whitelist now includes `devin`, so persisted Devin model selections are no longer silently rewritten to `codex`.
- **Fail-closed provider binding** (`apps/server/src/provider/Layers/ProviderService.ts`) — `startSession` no longer defaults an omitted provider to `codex`; it either derives the provider from the supplied `modelSelection` or returns a validation error.
- **First-turn provider precedence** (`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts`) — when writing the optimistic placeholder session row and when `ensureSessionForThread` validates a requested switch, the explicit `thread.turn-start-requested` model selection wins on a new/unbound thread, and only an established runtime binding is treated as immutable.
- **Turn settlement drain** (`apps/server/src/provider/acp/AcpAdapterSessionSupport.ts`, `DevinAdapter.ts`, `DroidAdapter.ts`, `GrokAdapter.ts`) — `DevinAdapter` now waits for all in-flight `session/update` notifications to be processed before emitting `turn.completed`, using a shared helper that Droid/Grok were already using.
- **Dead-session stop** — `DevinAdapter` stops the ACP session on prompt failure instead of reusing a closed child.
- **Process cleanup** (`apps/server/src/provider/acp/acpProcessCleanup.ts`) — tracked ACP children are killed on coordinated `Runtime.Teardown` (POSIX `SIGTERM`/`pkill`, Windows `taskkill /T`).
- **Atomic ACP session epoch transition & startup interaction registry** — `AcpSessionRuntime` installs sessions atomically with buffered notification replay; `requestPermission`/`elicitation` dispatches during startup go through a single `Ref` state machine.
- **Authoritative Devin configuration** — `applyDevinSessionConfiguration` resolves allowed models, calls `setModel`, re-reads `configOptions`, applies traits in a fixed order with dedup, and verifies postconditions.

Web/composer:
- **Devin-aware draft layer** (`apps/web/src/composerDraftModels.ts`) — `COMPOSER_PROVIDER_KINDS` and `normalizeProviderModelOptions` now include `devin`, so persisted draft options and provider detection stop silently dropping Devin selections.
- **Model picker** (`apps/web/src/components/chat/ProviderModelPicker.tsx`) — Devin model search is enabled and models are grouped with trait chips.
- **Composer file attachments** (`apps/web/src/components/chat/ComposerExtrasMenu.tsx`, `ChatView.tsx`, `composerSend.ts`) — the `+` menu accepts non-image file attachments when the provider supports them (Devin); selected files are routed to the image or generic-file attachment path as appropriate.
- **Settings migration** — `devinBinaryPath` and `customDevinModels` migrate to server settings on first load.
- **Thread title fallback** (`ProviderCommandReactor.ts`) — the configured title-generation model is skipped when its provider reports unavailable, falling back to a prompt-derived title.

Tests added:
- `modelSelectionCompatibility.test.ts` — Devin persistence/label inference.
- `ProviderCommandReactor.test.ts` — first-turn Devin binding, idempotent turn starts, provider-switch rejection.
- `AcpJsonRpcConnection.test.ts` — session resume and restart+resume.
- `DevinAdapter.test.ts` — two consecutive turns, restart/resume, configuration.
- `AcpAdapterSessionSupport.test.ts` — shared queued-turn-event drain helper.
- `composerSend.test.ts` and `ComposerExtrasMenu.browser.tsx` — mixed image/text file picker routing.

## Why

Review passes and live UI/UX testing flagged P0/P1 correctness gaps that made Devin threads unusable on a fresh state: persisted model selections being rewritten to Codex, new threads binding to Codex even when Devin was selected, second/follow-up turns hanging because `turn.completed` was emitted before queued ACP notifications drained, `.txt` attachments rejected in the composer, and the model picker lacking search. This revision fixes those while staying on the official ACP SDK and the shared runtime shape.

## UI Changes

Settings → Providers now shows the Devin install card (binary path, version, auth state). The composer model picker groups all Devin models by upstream lab and exposes Speed/Context/Effort/Thinking trait chips where applicable. The composer `+` attachment menu accepts text and other non-image files for Devin. Live UI/UX screenshots are captured in the gallery comments below.

## Checklist

- [x] This PR is scoped to the Devin ACP provider + required ACP runtime support
- [x] I included before/after screenshots for UI changes (see gallery comments)

## Validation

- [x] `bun fmt` — clean
- [x] `bun lint` — 0 errors (256 pre-existing warnings)
- [x] `bunx turbo run typecheck --filter=!@synara/marketing` — 7/7 packages pass
- [x] `apps/web` targeted tests — `composerSend.test.ts` 20/20 pass
- [x] Sub-agent/CI targeted reports — `modelSelectionCompatibility` 11/11, `ProviderCommandReactor` 102/102, `ProviderService` 65/65, `AcpJsonRpcConnection`, `DevinAdapter`, `AcpAdapterSessionSupport`, `ProviderModelPicker` browser tests pass

Note: the full `apps/server` test suite requires Node ≥ 24 for `node:sqlite`; `bun` on this VM does not expose `node:sqlite`, so local server tests were validated by a sub-agent running Node 24 and by CI.

## Stacked PRs

The earlier stacked verification PRs (#50 CLI health probe, #51 unit tests, #52 adapter integration tests) were closed and their contents merged into this single merge-ready PR. Only PR #1 is intended for upstream; upstream mirror is https://github.com/Emanuele-web04/synara/pull/268.

Link to Devin session: https://app.devin.ai/sessions/a3b6f650440f45069ef979ba8102931f
Requested by: @kartikkabadi


kartikkabadi:
@coderabbitai @cubic-dev-ai full review this whole PR, gather as much context as you can, compare it with the issues/other PRs, compare the diffs, compare the code, architecture, quality, implementation, and others like a senior engineer and do a full extensive scrutiny review.

kartikkabadi (commented):
## Review #1 — Authentication fallback can select an unusable method

**Head reviewed:** `e9817e819697376581dc86e8b62ec964fef906c8`

**Review scope:** Drafted as an early warning from a dedicated HUGE-PR review pass. I inspected the PR metadata and prior discussion, all 114 changed-file names, exact-head CI, and the ACP process/authentication trust boundary in `acpSpawnEnv.ts`, `DevinAcpSupport.ts`, `DevinAcpSupport.test.ts`, and the on-demand authentication flow in `AcpSessionRuntime.ts`. The remaining provider adapter, session/event ingestion, plan mode, model catalog, checkpoint, web UI, contracts, and compatibility surfaces have not yet received a complete second-pass review.

**Verdict:** PARTIAL—DO NOT RELY ON THIS YET

### Blocking findings

#### P1 — On-demand recovery prefers API-key auth even when no API key exists

`resolveDevinAcpAuthMethodId()` selects `windsurf-api-key` whenever Devin advertises it, before considering `browser_login`. That choice is independent of `hasDevinApiKeyEnv()`. In the on-demand flow, this resolver is called only after `session/new` or `session/load` reports authentication is required.

This violates the stated recovery invariant: users relying on stored `devin auth login` credentials should be able to recover through browser login when those credentials are missing or expired. When both methods are advertised but `WINDSURF_API_KEY` is unset, the runtime still sends `authenticate({ methodId: "windsurf-api-key" })`; the browser method is never attempted. A first-time or expired-credential user can therefore be unable to start Devin despite a valid browser-login method being available.

**Minimal fix:** make method selection conditional on credential availability: choose `windsurf-api-key` only when `hasDevinApiKeyEnv()` is true, otherwise choose `browser_login`. If API-key authentication itself fails and browser login is advertised, consider one explicit fallback rather than failing the whole start.

**Verification:** add a test where both methods are advertised and `WINDSURF_API_KEY` is absent; assert `browser_login` is selected. Add the inverse test with a non-empty key, plus an integration-style on-demand test proving an auth-required `session/new` retries through browser authentication.

**Confidence:** High. The resolver, tests, and runtime call path all encode the unconditional API-key preference directly.

### Coverage matrix

| Area | Evidence inspected | Confidence | Residual risk |
|---|---|---:|---|
| PR intent / state | PR body, head/base metadata, prior comments | High | Mergeability currently reports false and was not independently diagnosed |
| Exact-head CI | CI run `28972474056` | High | CI is green, but live Devin/browser authentication is environment-dependent |
| Child-process environment | `acpSpawnEnv.ts`, Devin spawn construction and tests | High | Other provider-specific environment paths remain unreviewed |
| ACP authentication | resolver, tests, on-demand retry path in `AcpSessionRuntime.ts` | High | Authentication behavior against the live Devin CLI was not reproduced here |
| Remaining server/provider implementation | Changed-file inventory only | Low | Full adapter, lifecycle, event, model, mode, elicitation, and checkpoint review remains |
| Web/UI/contracts | Changed-file inventory only | Low | Full behavior and compatibility review remains |

### Validation and residual risk

The exact head has a successful CI workflow. I did not execute the repository or the live Devin CLI. Because this PR is approximately 12,800 additions across 114 files and material subsystems remain unexamined, this is intentionally not a complete review. The finding above warrants early warning; the PR still needs a full risk-sliced follow-up before this review can be treated as complete.

Review #1
Reviewed by ChatGPT - Sol

kartikkabadi:
### Automated review disclosure

All reviews and inline comments in this PR thread signed `Reviewed by ChatGPT - Sol` were generated and posted automatically by Kartik’s GitHub Review Watch automation. They were not manually authored or manually submitted by Kartik.

This disclosure applies retroactively to **Review #1** and any associated inline comments. It does not change the technical findings, verdict, reviewed SHA, or review status.

Review #1
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #2 — Follow-up

**Head reviewed:** `19c2ee0a9a66cf69ddbe2076e9dbd165f7778f57`

**Verdict:** NO MATERIAL FINDINGS

I rechecked the previously identified authentication path and the later elicitation-regex hardening. Authentication now selects API-key mode only when the matching credential exists and otherwise permits browser login. Regex validation now executes within an interruptible Node VM timeout and fails closed on timeout or execution error.

I also reviewed the surrounding ACP runtime, process-environment boundary, rollback capability gate, provider adapter, model and mode mapping, event ingestion, and relevant tests. Exact-head GitHub Actions CI is green. Remaining risk is primarily live integration behavior that was not independently reproduced here; I found no concrete correctness, security, or data-integrity defect at this head.

Review #2
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #3 — Follow-up

**Head reviewed:** `adbc563657ba3185405d19cc9529793979858fe4`

**Previous reviewed head:** `19c2ee0a9a66cf69ddbe2076e9dbd165f7778f57`

**Verdict:** NO MATERIAL FINDINGS

I reviewed the intervening provider/runtime changes and rechecked the prior high-risk boundaries.

The rollback contract is now internally consistent: Devin no longer advertises conversation rollback that its adapter cannot perform, and checkpoint revert fails closed before mutating the filesystem when provider rollback is unavailable. The temporary Droid settings regression introduced during provider wiring is also repaired across provider visibility, dirty-state computation, displayed binary path, and persisted settings updates.

ACP child-process cleanup is now integrated with the Effect runtime teardown instead of installing an independent signal owner. This preserves the application's coordinated asynchronous shutdown path while synchronously removing tracked ACP descendants before the runtime exits. The accompanying subprocess regression coverage exercises the termination boundary.

I also rechecked the later live-diff coalescing change, which now scopes pending diffs by both thread and turn, preventing events from separate turns from being merged together.

GitHub currently exposes no pull-request workflow run for this exact head in the upstream repository, so this review does not claim independent exact-head CI evidence there. The same commit is present on the fork PR and its targeted validation was previously reported green. I found no concrete correctness, security, compatibility, or data-integrity defect in the reviewed delta.

Review #3
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #2 — authentication blocker fixed; huge-PR review remains incomplete

**Head reviewed:** `adbc563657ba3185405d19cc9529793979858fe4`

**Scope:** focused follow-up on the Review #1 blocker and the adjacent ACP shutdown/authentication changes. This is not a complete second-pass review of the full ~5.3k-line provider port.

### Prior blocker status

The Review #1 authentication defect is fixed at this head. `resolveDevinAcpAuthMethodId(...)` now checks `hasDevinApiKeyEnv(env)` before selecting `windsurf-api-key`, prefers `browser_login` when no key is present, and only falls back to the API-key method when that credential is actually available. This matches the requested invariant and removes the unusable-method selection path identified at the previous SHA.

The added process-cleanup path is also materially better than the earlier implementation shape: tracked ACP child PIDs are killed from the runtime teardown after Effect finalizers complete, and the subprocess regression fixture explicitly verifies that coordinated asynchronous cleanup is not preempted.

### Verdict

**No new material finding in the reviewed slice.** Review #1's P1 is resolved.

However, this PR is still too large to treat this focused pass as a complete review. The remaining high-risk surfaces still requiring a dedicated pass include:

- `DevinAdapterLive` session ownership, cancellation, and event-ordering behavior;
- approval and elicitation request correlation;
- session resume/load semantics and stale runtime cleanup;
- model discovery/variant mapping against malformed or changing CLI output;
- web settings/model-picker compatibility and provider migration behavior;
- exact-head CI, which was not exposed by the connected workflow-run endpoint for this SHA.

Do not interpret this comment as merge readiness. It only closes the previously reported authentication blocker and records the reviewed shutdown slice.

Review #2
Automated
Reviewed by ChatGPT - Sol

kartikkabadi:
All review-gate items from the fork PR review (kartikkabadi/synara#1, comment 5019409387) are now resolved on head `4ed5f62282a9852a4f43e55e73185a4ac2f476a5`:

- `upstream/main` merged into `devin-acp-provider` with a merge commit; this PR is now mergeable with no conflicts.
- Droid model-catalog regression in `useProviderModelCatalog.ts` fixed.
- Cross-provider regression tests added (`apps/web/src/hooks/useProviderModelCatalog.browser.tsx`).
- Stacked fork PRs #50, #51, and #52 merged into `devin-acp-provider` and closed.
- `devin` target option rules added to `apps/server/src/agentGateway/targetResolver.ts`.
- `bun fmt`, `bun lint`, `bunx turbo run typecheck`, `apps/web` build, and targeted provider/browser regression tests pass.

The full local `apps/server` test suite has pre-existing Bun/Node 20.18.1 environment failures that do not reproduce in CI.

kartikkabadi (commented):
## Review #4 — Devin agent-gateway target rules

**Head reviewed:** `4ed5f62282a9852a4f43e55e73185a4ac2f476a5`

**Previous reviewed head:** `adbc563657ba3185405d19cc9529793979858fe4`

**Verdict:** NO MATERIAL FINDINGS IN THE NEW DEVIN-SPECIFIC DELTA

I reviewed the new `devin` entry in the agent-gateway target-option registry against the existing resolver semantics and the provider model descriptor contract.

The mapping is internally consistent with the Devin catalog shape:

- `reasoningEffort` is the primary string option and is constrained by model discovery;
- `fastMode` and `thinking` remain non-advertised unless the selected model exposes the corresponding capability;
- `contextWindow` is validated against the model descriptor rather than accepting arbitrary values;
- unknown option keys continue to fail closed through the shared resolver.

I found no concrete correctness, security, or data-integrity defect in this commit. However, the registry-wide target-resolver test matrix at this head does not yet include a Devin case, so this review does not claim direct regression coverage for accepted and rejected Devin option combinations. Adding Devin cases for reasoning effort, fast mode, thinking, and context window would materially strengthen the proof, but the absence of those tests alone is not treated as a blocking defect here.

No exact-head pull-request check/status was exposed for this SHA through the connected GitHub endpoints. This remains a focused huge-PR follow-up, not a claim that every intervening merged-main change or the entire provider port has been exhaustively re-reviewed.

Review #4
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #5 — Devin adapter hardening follow-up

**Head reviewed:** `472eadf88182b4411bb7514c4d143d8347004e4e`

**Previous reviewed head:** `4ed5f62282a9852a4f43e55e73185a4ac2f476a5`

I reviewed the seven-commit delta between these heads, covering the changed Devin adapter, elicitation, model-catalog, ACP process-cleanup, and prompt-attachment surfaces, plus the exact-head CI result.

The delta materially improves failure typing and recovery behavior: ACP discovery failures are normalized into provider-adapter errors, interrupted turns now run the intended cleanup finalizer, preflight model/mode failures are tested not to wedge `activeTurnId`, model discovery/parser coverage is expanded, and attachment traversal validation is exercised with an actually invalid identifier. The exact head has a successful CI run.

I found no concrete correctness, security, lifecycle, or data-integrity defect in this delta. This remains a focused huge-PR follow-up rather than an exhaustive re-review of the full Devin provider port; unreviewed residual risk remains in the broader adapter/event-ingestion and live Devin CLI behavior.

Review #5
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #6 — Provider discovery and model-normalization follow-up

**Head reviewed:** `65aa7ad92e7aa56b504d41706a66ad2f8c9e3b65`

**Previous reviewed head:** `472eadf88182b4411bb7514c4d143d8347004e4e`

I reviewed the seven-commit delta between these heads, covering provider-health behavior, Devin adapter/model discovery, ACP process cleanup, model capability propagation into the web client, provider settings cleanup, target resolution, and shared model-slug normalization.

The latest alias-normalization fix is internally consistent: alias lookup is now case-insensitive while unknown non-alias model identifiers preserve their original spelling, and the added tests cover multiple providers. The surrounding delta also adds substantial provider-health regression coverage and keeps capability-driven UI behavior tied to discovered model descriptors rather than hard-coded provider assumptions.

Exact-head GitHub Actions CI completed successfully. I found no concrete correctness, security, lifecycle, compatibility, or data-integrity defect in the reviewed delta.

This remains a focused huge-PR follow-up, not an exhaustive re-review of the entire Devin provider port or a substitute for live Devin CLI validation.

Review #6
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #7 — Fresh-session update isolation regression

**Head reviewed:** `7aad2304fbfe83f5b8827abbeef625229efc10c0`

**Previous reviewed head:** `65aa7ad92e7aa56b504d41706a66ad2f8c9e3b65`

**Verdict:** MATERIAL FINDING

The force-pushed reset substantially reduces the PR and exact-head CI is green, but the new ACP session-epoch logic still admits updates from the wrong session during an ordinary successful `session/new` call.

### P1 — `PENDING_SESSION_ID` temporarily treats every session ID as active

For a fresh session, `runSessionSetup` sets the active epoch to `PENDING_SESSION_ID` before awaiting `session/new`. `isActiveSessionId(...)` then returns true for **any** notification session ID while that sentinel is active. Those notifications can mutate bounded state and enqueue transcript/tool events.

When `session/new` succeeds normally, the runtime switches the epoch to the returned session ID, but it does not reconcile or drain events admitted under the pending wildcard. The cleanup path only runs for a discarded unauthenticated probe session. Therefore, a delayed notification from an older ACP session—or any unrelated session update arriving while `session/new` is in flight—can remain in the new Synara session's event queue and state.

This is especially risky for long-lived or reconnecting ACP processes where stale notifications are plausible. The current mock coverage proves cleanup for the auth-retry probe, but not isolation for a successful fresh-session setup receiving a mismatched session update.

**Minimal fix:** buffer pending-session notifications keyed by their actual `notification.sessionId`, then retain/apply only the bucket matching the `sessionId` returned by `session/new`; discard every other bucket before enabling normal delivery. Do not use a wildcard active-session predicate that directly mutates shared state or the canonical event queue.

**Verification:** add a test where `session/new` emits an update for `old-session` before returning `new-session`, then assert that no command/config/transcript/tool state from `old-session` is observable after startup. Also preserve the existing early-update test for the eventual returned session ID.

Exact-head GitHub Actions CI completed successfully, but the current tests do not exercise this mismatched-success path.

Review #7
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #8 — Pending-session isolation is fixed, but the new buffer is unbounded

**Head reviewed:** `2174d84e9bf98c1cbab2690b1cd41610ccc48a3d`

**Previous reviewed head:** `7aad2304fbfe83f5b8827abbeef625229efc10c0`

**Verdict:** MATERIAL FINDING

The Review #7 cross-session contamination path is fixed: notifications received before `session/new` returns are now keyed by their claimed session ID, and only the bucket matching the authoritative returned session is replayed. The auth-probe regression test also demonstrates that orphan transcript and bounded state do not leak into the final session.

### P1 — Pre-session notifications now bypass the runtime's bounded queue

While `activeSessionId` is unset, every incoming `session/update` is appended to `pendingSessionUpdatesRef` using an unbounded array per arbitrary session ID:

```ts
const list = next.get(sessionId) ?? [];
next.set(sessionId, [...list, notification]);
```

This path has neither a total item/byte limit nor a per-session/session-count limit. It also copies the entire accumulated array for every notification, producing quadratic allocation work for a large bucket. The normal canonical event path is deliberately bounded at 2,048 events, but the new pre-session path bypasses that protection until `session/new` resolves or times out.

A malfunctioning, compromised, or merely noisy ACP child can therefore emit a high-rate stream of updates—or continuously vary `sessionId`—while session setup is in flight and drive unbounded memory/CPU growth in the Synara server. The 8 MiB frame guard limits individual frames, not aggregate buffered data.

**Minimal fix:** place a strict aggregate budget on pending updates before storing them. Prefer a bounded queue/ring or explicit item-and-byte accounting with a small cap; reject/terminate the ACP runtime once exceeded rather than silently dropping transcript events. Also cap the number of distinct pending session IDs. Avoid repeated `[...list, notification]` copies for incremental accumulation.

**Verification:** add a test that withholds the `session/new` response, emits more than the configured pending budget (including many distinct fake session IDs), and asserts startup fails closed with a typed transport/protocol error while process memory remains bounded. Preserve the existing test proving early updates for the eventual authoritative session are replayed.

No exact-head pull-request workflow run was exposed through the connected GitHub endpoint for this SHA, so this review does not claim independent exact-head CI evidence.

Review #8
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Automated review — blocking findings

I reviewed the exact-head delta after the prior upstream review, first revalidating the discarded-session race, then checking the replacement pending-session protocol, startup interactions, Plan enforcement, configuration authority, discovery scope, and exact-head CI. The branch architecture is substantially better and should remain lean, but it is not merge-ready.

### P1 — Pre-session update buffering is unbounded and can deadlock startup

While `session/new` is pending, every notification is copied into `pendingSessionUpdatesRef` with no count, byte, per-session, or global bound. A buggy or hostile ACP child can therefore consume unbounded memory and quadratic copying work before an authoritative session exists.

There is also a deterministic startup deadlock: after `session/new` returns, `setSessionEpoch()` replays buffered notifications into the bounded 2,048-entry event queue before `DevinAdapter.startSession()` has attached its event consumer. More than the queue capacity of valid early events causes replay to block forever, so startup never returns and the consumer is never created.

Use a single bounded pending-session state machine. Install the setup baseline, atomically select the final session id and detach its bounded pending data, discard all other ids, then reduce pending updates in order through the same reducer as live updates. Replay must not depend on an event consumer that cannot exist until startup completes. Add process-backed tests for pending limits and an early stream larger than the event-queue capacity.

### P1 — Early authoritative updates can be overwritten by the older setup response

The transition commits/replays pending state before applying `sessionSetupResult.modes` and `sessionSetupResult.configOptions`. A newer early mode/config notification can therefore be reduced and then overwritten by the older response snapshot.

Apply the setup response as the baseline first, then replay matching pending notifications in arrival order. The test should prove final mode, full config, commands, tool state, assistant state, and transcript state—not only absence of one orphan text chunk.

### P1 — Startup approvals and elicitation cannot be answered

Approval and elicitation handlers are registered before `acp.start()`, but the adapter does not add the session to its routable `sessions` map until startup and configuration finish. If Devin emits an approval or elicitation request during authentication or session creation, Synara can publish it and then wait indefinitely while `respondToRequest` / `respondToUserInput` reject the response because the session is not registered.

Introduce a generation-scoped provisional startup owner (or startup-interaction registry), promote it atomically on success, and clear it on failure, cancellation, replacement, or stale generation. Add process-backed startup approval and elicitation tests proving the response reaches the waiting ACP handler.

### P1 — Plan still fails open when mode state is absent

`resolveRequestedModeId()` returns success when `modeState` is undefined. For `interactionMode === "plan"`, that permits `session/prompt` with only textual instructions and no verified safe provider mode. Prompt text is not an enforceable no-mutation boundary.

Plan must reject before prompt dispatch when mode state is absent, no deliberate safe alias exists, the switch fails, or the resulting current mode cannot be verified. If current Devin ACP does not expose an enforceable Plan/Ask mode, disable Plan for Devin. Add the missing `modeState: undefined` regression and assert no prompt request is sent.

### Additional correctness gaps

- Mode matching is exact lowercase equality despite the PR calling it whole-token matching. Validate current Devin IDs/names and normalize case, whitespace, `_`, and `-` while allowing only deliberate aliases.
- Configuration deduplicates two logical traits targeting one ACP selector without detecting conflicting requested values; it also validates case-insensitively but sends the caller spelling, and does not verify each post-mutation `currentValue`.
- The generic ACP runtime treats an empty model selector as proof authentication is required. Fixed-model/no-model-selector agents can be valid; this heuristic should be Devin-specific rather than a generic ACP invariant.
- Opening any model picker can still trigger Devin discovery, and Devin prefetch omits `cwd`, producing mismatched cache keys and potentially duplicate processes in different working directories.
- Static fallback reports success without retaining a structured, redacted reason live discovery failed.

Exact-head CI is green, but it does not cover the startup deadlock, startup interaction routing, or fail-open Plan paths above.

Review #4
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #9 — Startup interaction registry is not generation-safe

**Head reviewed:** `0c8791b0508dd0f796f637b95c48325c03145aa7`

**Previous reviewed head:** `2174d84e9bf98c1cbab2690b1cd41610ccc48a3d`

**Verdict:** MATERIAL FINDING

The prior pending-session findings are materially addressed: buffering is now bounded, setup state is installed before matching notifications are reduced, transcript delivery no longer blocks startup on the event queue, startup approval/elicitation routing exists, Plan fails closed when mode state is absent, and discovery/configuration postconditions are stricter.

### P1 — Concurrent startup requests can be lost or cross into a later startup generation

`makeStartupInteractionRegistry.dispatch()` performs a non-atomic read/modify/write of `pendingRef`:

```ts
const pending = yield* Ref.get(pendingRef);
...
yield* Ref.set(pendingRef, pending.concat({ req, deferred, generation }));
```

Two approval or elicitation requests arriving concurrently can both read the same array and then overwrite each other. The overwritten request's deferred is no longer reachable by `flush()` or `cancel()`, so the ACP peer can wait forever for a response.

There is a second race at the startup boundary. `dispatch()` reads `generationRef`, but `flush()` and `cancel()` never check the stored generation. If `begin()` increments the generation and cancels while a dispatch fiber is between `Ref.get(pendingRef)` and `Ref.set(...)`, that stale request can be appended after cancellation. It may then be flushed through the next startup's handler, crossing session generations, or remain suspended indefinitely if startup fails.

The `generation` field currently records the intended fence but does not enforce it.

**Minimal fix:** make admission atomic with `Ref.modify` (or a queue/STM structure) so capacity checking and insertion happen in one operation. During admission, compare the current generation and lifecycle state in the same atomic transition. `begin`, `complete`, and `cancel` must atomically detach only the applicable generation, and stale entries must complete with the safe default rather than reaching a later handler.

**Verification:** add adversarial tests that synchronize two dispatch fibers at admission and prove both complete exactly once; then interleave `dispatch` with `begin()`/`cancel()` and assert the old-generation request receives the safe default and is never observed by the new handler. Sequential tests do not exercise this race.

No pull-request workflow run was exposed for this exact head through the connected GitHub endpoint, so this review does not claim independent exact-head CI evidence.

This remains a focused follow-up on the newest 16-file delta, not an exhaustive re-review of every file in the full provider port.

Review #9
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Follow-up review — no material findings

**Head reviewed:** `69aa10c5fdb8248c33d2dc95c4e4e96bd4442025`

**Verdict:** NO MATERIAL FINDINGS

This pass reviewed the current effective PR surface after the startup/session hardening series and revalidated the previously blocking areas:

- startup notifications are bounded and cannot deadlock session creation before the consumer exists;
- the final session baseline is installed before buffered notifications are replayed, so newer authoritative mode/config updates are not overwritten by the older `session/new` response;
- startup permission and elicitation requests are owned by an atomic, generation-scoped registry and cannot become stranded during registration races;
- on-demand authentication discards the unauthenticated probe session without allowing its notifications or state to cross into the authenticated session;
- every session update is fenced by the exact session id and monotonic generation through the final state-mutation and queue boundaries;
- Devin Plan mode fails closed unless an exact enforceable ACP mode is available;
- requested model and trait configuration is verified against provider-confirmed postconditions;
- authentication retry classification is provider-scoped rather than inferred from arbitrary failures;
- model discovery is scoped to Devin selection, includes the working directory where required, and preserves truthful fallback diagnostics;
- ACP child cleanup remains integrated with coordinated runtime teardown rather than competing signal handlers.

The exact reviewed head has a successful upstream CI run. No new correctness, security, lifecycle, or compatibility blocker was found in this revision. Residual risk is primarily live Devin CLI behavior that cannot be fully reproduced by repository tests alone; the implementation now fails closed at the important trust boundaries.

Review #5
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #10 — The branch has absorbed a large unrelated cleanup stack

**Head reviewed:** `961e45092aa9cb932b53bb18add7ab602864e423`

**Previous reviewed head:** `0c8791b0508dd0f796f637b95c48325c03145aa7`

**Verdict:** BLOCKED ON SCOPE / HISTORY CORRECTION

The newest head is **33 commits ahead** of the previously reviewed Devin-provider head. That delta is not a focused fix for Review #9. It includes a broad repository cleanup/refactor stack spanning desktop browser/session policy, resumable updates, Codex/Claude/OpenCode adapters, provider runtime ingestion, persistence projections, terminal management, ChatView decomposition, settings, draft stores, and many other unrelated surfaces.

Concrete examples in the delta include:

- removing `apps/server/src/attachmentUpload.ts` and its tests;
- moving roughly 1,000 lines out of `ProviderRuntimeIngestion.ts` into a new projection module;
- large rewrites of `codexAppServerManager.ts`, `OpenCodeAdapter.ts`, `ChatView.tsx`, browser management, terminal management, and multiple persistence layers;
- adding broad cleanup/audit work unrelated to the Devin ACP provider;
- the current tip itself is a generic draft-model fallback refactor shared by ChatView and DiffPanel.

This makes the PR materially different from its stated contract and invalidates the current size/scope declaration. The body still says **55 files, +4,005 / −307 at `0c8791b0`**, while the branch now contains a large additional cleanup stack. Even though exact-head CI is green, CI cannot establish that these unrelated architectural changes belong in this PR or that the full combined history has received an adequate review.

### Required correction

Restore this branch to a Devin-only history. The safest path is to reconstruct or reset the PR branch from the last intended Devin head, then reapply only the commits that directly fix Review #9 and any other Devin-specific findings. Do not retain the unrelated cleanup stack merely because it is green.

After correction:

1. update the PR body with the exact new head, file count, and diff size;
2. provide a commit list showing that every retained commit is necessary for the Devin ACP provider;
3. rerun exact-head CI;
4. preserve the existing review history rather than opening a replacement PR unless branch repair is impossible.

I am not attempting a technical review of the 33-commit cleanup stack inside this provider PR. Until the branch is narrowed again, further Devin-specific review would be low-confidence because unrelated refactors can change the same runtime, adapter, persistence, and UI surfaces under review.

Review #10
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Follow-up review: no material findings

I reviewed the four-commit delta from `69aa10c5` to `961e4509`, limited to the ten changed web files rather than re-reviewing the unchanged ACP runtime.

The delta consistently centralizes draft-thread model fallback selection across ChatView and DiffPanel, preserves explicit project defaults ahead of the global default provider, skips `pi` where no canonical default model exists, and falls back through the provider-specific model table before Codex. The call sites now use the same normalized `ModelSelection`, including first-render diff previews, and the focused tests cover project/default-provider precedence and provider-specific fallbacks.

I also checked the adjacent first-send, diff-panel, and provider-settings changes in this delta for stale dependencies, mismatched draft/server-thread behavior, provider/model divergence, and regressions in existing fallback paths. I did not find a concrete correctness, security, or data-integrity defect at this head.

Exact-head CI completed successfully.

Review #6
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #11 — Scope blocker remains, and exact-head CI is red

**Head reviewed:** `4d1873c2902448232867c885ae8aa67f32faf3bc`

**Previous reviewed head:** `961e45092aa9cb932b53bb18add7ab602864e423`

**Verdict:** BLOCKED

The branch has advanced by another **18 commits** without rewriting away the unrelated cleanup history identified in Review #10. Because this head is strictly ahead of `961e4509`, the 33-commit cleanup stack is still part of the PR; the Devin-specific additions do not resolve the scope problem.

There is also a new exact-head CI failure. Format, lint, typecheck, and the non-browser test suite pass, but the stable browser-test step fails three tests. Two failures are directly in the newly changed attachment surface:

- `ComposerExtrasMenu.browser.tsx > routes picked images and generic files to their composer callbacks`
- `ComposerExtrasMenu.browser.tsx > shows the attachment action in the menu`

The test helper defaults `supportsFileAttachments` to `false`, while both rewritten assertions expect the `true` branch (`accept === ""` and `Add attachment`). That makes the new tests internally inconsistent with their own fixture. The third failure is in the existing ChatView settled-layout timing test.

### Required next steps

1. Repair the branch history so this PR contains only the Devin provider and genuinely required shared-runtime support. Adding more commits on top does not remove the unrelated cleanup stack.
2. Fix the attachment browser-test fixture or explicitly mount with `supportsFileAttachments: true` for the generic-file assertions, while retaining a separate image-only assertion for unsupported providers.
3. Investigate the ChatView browser failure and establish whether it is deterministic or runner timing noise.
4. Rerun the full exact-head CI after both scope repair and test correction.

I am not continuing a broad technical pass while the PR still contains the unrelated stack and its exact-head required CI is failing. Both conditions materially reduce review confidence.

Review #11
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Automated review — follow-up

The branch has moved substantially since Review #6. I rechecked the effective Devin/ACP integration surface, including first-turn provider binding, thread-scoped MCP injection, model/variant discovery, attachment routing, authentication retry isolation, session-generation fencing, queued ACP event draining, configuration postconditions, provider health fallback for title generation, and the associated server/web regressions.

I found no new material correctness or security defect in the reviewed code delta.

This head is **not merge-ready yet**, because exact-head CI is red. Format, lint, typecheck, unit tests, release smoke, and the Windows process regression passed, but the stable Chromium browser suite failed in `ChatView.browser.tsx` at `keeps the live inline-tool layout through the first settled paint, then relaxes after the grace delay` (`expected null not to be null`). The desktop build and preload verification were consequently skipped. Re-run or fix that browser regression and require a fully green exact-head run before merge.

Review #7
Automated
Reviewed by ChatGPT - Sol

kartikkabadi:
Quick note on the `Format, Lint, Typecheck, Test, Browser Test, Build` failure: it is a single flaky `ChatView.browser.tsx` timeline-estimator timeout (`keeps the live inline-tool layout through the first settled paint...`). The same commit passes that full suite in the fork PR (#1) and the test passes locally on two consecutive runs, so this appears to be a runner/timing flake rather than a code defect. I do not have permission to retrigger the upstream workflow run; a rerun should clear it.

kartikkabadi (commented):
## Review #12 — Scope repair and runtime race fixes verified

**Head reviewed:** `d4075969ef724703aa9ca2dce4594c22e67a40e9`

**Previous reviewed head:** `4d1873c2902448232867c885ae8aa67f32faf3bc`

**Verdict:** PRIOR BLOCKERS RESOLVED; HUGE-PR REVIEW REMAINS PARTIAL

The branch history has been repaired. The unrelated repository-wide cleanup stack reported in Reviews #10 and #11 is no longer present in the PR diff. The current changed-file inventory is again confined to the Devin provider, shared ACP runtime support, provider/model contracts, composer/provider UI, and their tests.

Exact-head CI run `29790166217` completed successfully, so the attachment-browser and ChatView failures reported in Review #11 are no longer active at this head.

I also rechecked the two ACP-runtime defects previously reported:

1. **Session epoch transition / orphan update isolation:** the runtime now uses a bounded, session-keyed pending-notification state machine and includes a deterministic transition-window regression test proving a matching `session/update` is applied exactly once while the epoch changes, with no pending state left behind.
2. **Startup approval/elicitation concurrency:** the startup interaction registry now performs dispatch, begin, register, complete, and cancellation transitions through atomic `Ref.modify` operations; pending entries carry a generation fence, stale entries receive the safe default, the buffer is bounded, and tests cover concurrent dispatches, register/complete races, stale generations, cancellation, exactly-once delivery, and overflow.

I did not find a new concrete correctness, security, lifecycle, or data-integrity defect in this focused verification pass. This does not convert the review into a complete approval: the PR remains a large cross-layer change, and the still-unreviewed portions of the provider adapter, model/configuration semantics, persistence compatibility, and web interaction paths retain residual risk.

Review #12
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #8 — No new material findings after upstream synchronization

**Head reviewed:** `d4075969ef724703aa9ca2dce4594c22e67a40e9`

**Previous reviewed head:** `47e6d07b48f2f299458c7a86ac3556fced2664ad`

**Delta:** 12 commits. This is a large synchronization delta spanning Agent Gateway diagnostics and managed worktrees, ACP runtime consolidation, provider adapters, orchestration persistence, WebSocket admission/transport, and web timeline/sidebar changes. I treated it as a huge-PR follow-up and reviewed it by risk slice rather than assuming the earlier verdict carried forward.

**Verdict:** NO NEW MATERIAL FINDINGS

The previously reviewed Devin-specific invariants remain intact at this head:

- authentication and replacement-session transitions remain fenced;
- startup notifications and interactions remain bounded and session-scoped;
- model/mode configuration remains fail-closed and postcondition-checked;
- provider child-process environment and coordinated teardown boundaries remain preserved;
- turn completion remains ordered after queued provider events;
- explicit provider/model selection is not silently rebound to Codex;
- generic file attachments remain constrained to providers that advertise support.

The new synchronization work also removes the generated `effect-acp` workspace in favor of the shared ACP SDK/runtime path, adds operational diagnostics and cursoring, introduces managed-worktree support, and extends stream admission and timeline grouping. I did not find a concrete correctness, security, lifecycle, or compatibility defect in those changes that meets the protocol's reporting threshold.

The exact-head GitHub Actions CI run completed successfully. Residual risk remains higher than for a focused PR because this synchronization touches a broad cross-section of the application; this review should not be interpreted as endorsing the PR's overall size or maintainability.

Review #8
Automated
Reviewed by ChatGPT - Sol

kartikkabadi:
Superseded by #772 — rebased integration of this PR with #584 (sebbonit's), on current main, verified merge-ready. All ACP runtime hardening, fail-closed behaviors, and tests from here are carried in; model mechanism switched to the documented `--model` + variant-UID path. Thanks @sebbonit — usage tracking, log redaction, compaction, slash discovery, and the variant matrix all landed via your branch.

PR #274, opened 2026-06-30
fix(claude): keep macOS Claude OAuth token fresh to stop ~8h 'not logged in'

Faxraddin:
## Problem

On macOS, the Claude provider intermittently reports **"not logged in"** roughly every ~8 hours, even though `claude auth status` shows the user is logged in. Re-logging in fixes it for another ~8h.

**Root cause:** On macOS, Claude Code stores OAuth credentials in the login **Keychain** item `Claude Code-credentials` (`accessToken` + `refreshToken` + `expiresAt`), and the access token has an **~8h TTL**. The Claude auth path (`apps/server/src/provider/claudeProcessEnv.ts`) only inspects the **file** `~/.claude/.credentials.json`, which **does not exist on macOS**. So the expiry is never observed, no refresh is triggered, and a long-lived Claude Agent SDK session rides a token that lapses after ~8h.

(Confirmed empirically: a freshly issued token reports ~7.8h of remaining lifetime, matching the observed cycle.)

## Fix

A small, defensive background job that periodically runs the official **`claude` CLI** (`claude auth status`), which validates and refreshes its **own** Keychain token via its **own** Keychain ACL — so there's no auth prompt and no refresh-token-rotation logic to get wrong here. This keeps the Keychain token perpetually fresh, so the SDK session always reads a valid token.

Deliberately **not** reading/writing the Keychain or implementing the OAuth refresh in-app, to avoid Keychain prompts and refresh-token-rotation hazards — it delegates entirely to the official CLI.

- macOS-only (no-op elsewhere; other platforms use the file the SDK already manages)
- Never throws (best-effort; self-heals on the next tick)
- Opt out: `T3CODE_CLAUDE_KEEPALIVE=0`
- Tune interval: `T3CODE_CLAUDE_KEEPALIVE_MINUTES` (default 30)

## Files
- `apps/server/src/provider/claudeCredentialKeepalive.ts` — new background job
- `apps/server/src/main.ts` — 1 import + 1 startup hook next to `startServerMemoryDiagnostics`

## Notes
- The one thing not provable without waiting out a full ~8h cycle is that `claude auth status` performs the refresh at the exact near-expiry moment; it must validate the token to report status, and the job self-heals on the next tick if a cycle is missed. If preferred, the trigger could be swapped for a minimal `claude -p` ping.

### DIFF
--- a/apps/server/src/main.ts
+++ b/apps/server/src/main.ts
@@ -25,6 +25,7 @@ import { Open } from "./open";
 import * as SqlitePersistence from "./persistence/Layers/Sqlite";
 import { makeServerProviderLayer, makeServerRuntimeServicesLayer } from "./serverLayers";
 import { startServerMemoryDiagnostics } from "./memoryDiagnostics";
+import { startClaudeCredentialKeepalive } from "./provider/claudeCredentialKeepalive";
 import { ProjectionSnapshotQuery } from "./orchestration/Services/ProjectionSnapshotQuery";
 import { ProviderHealthLive } from "./provider/Layers/ProviderHealth";
 import { ProviderSessionReaperLive } from "./provider/Layers/ProviderSessionReaper";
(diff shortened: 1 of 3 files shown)


Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

PR #279, opened 2026-07-01
fix(grok): prevent ACP resume replay from OOMing the backend

### LINKED ISSUE #278: Grok ACP session/load replay can OOM the desktop backend before the event consumer attaches
## Summary

Resuming a Grok ACP session (`session/load`) can stream a large amount of `session/update` replay traffic before Synara attaches a consumer to the shared ACP runtime event queue. Because that queue is unbounded, a long Grok history replay can balloon process memory and crash the desktop backend (often around Electron’s ~2GB practical limit).

## Reproduction

1. Use Synara desktop with the Grok provider.
2. Resume a thread with a long Grok ACP history.
3. Observe backend memory climbing rapidly during session startup; the process may exit with OOM.

## Root cause

Two timing gaps stack together:

1. **ACP runtime:** `session/load` may emit many replay notifications while `AcpSessionRuntime.start()` is still running, before any adapter drains `getEvents()`.
2. **Grok adapter:** even after `start()` returns, the notification fiber was previously forked only *after* `applyRequestedSessionConfiguration()`, leaving another window where replay could accumulate.

Replay chunks are stale transcript history and must not be projected into the next live turn, but they still need to be drained or dropped cheaply.

## Proposed fix

- Drop `session/update` notifications in `AcpSessionRuntime` until `start()` completes (covers replay during `session/load`).
- In `GrokAdapter`, fork the ACP notification consumer immediately after `start()`, then apply session configuration, then wait for replay to go quiet before marking the session ready.
- Suppress any residual replay at the adapter layer and map Grok `/compact` to `context_compaction` UI rows.
- Pass both `--max-old-space-size` and `--js-flags=--max-old-space-size=...` to the packaged desktop backend child process.

## Impact

- Grok session resume stability
- Desktop backend memory usage during startup
- `/compact` support in the Grok provider

I have a patch ready in fork branch `chyax98:fix/grok-acp-resume-replay-oom`.

chyax98:
Fixes #278

## Problem

Grok `session/load` can replay a large history through ACP `session/update` notifications before Synara attaches an event consumer. The shared ACP runtime uses an unbounded queue, so resume startup could accumulate enough replay traffic to OOM the desktop backend.

## Approach

This is a layered fix:

1. **ACP runtime (`AcpSessionRuntime`)** — drop `session/update` notifications until `start()` finishes, so replay emitted during `session/load` never enters the queue.
2. **Grok adapter (`GrokAdapter`)** — fork the notification consumer immediately after `start()`, before `applyRequestedSessionConfiguration()`, so any post-start replay is drained while adapter-level suppression is active; then wait for replay to go quiet before exposing the session.
3. **Compaction** — implement `compactThread` via Grok `/compact` and map compaction tool calls to `context_compaction` runtime rows.
4. **Desktop backend heap** — pass both `--max-old-space-size` and `--js-flags=--max-old-space-size=...` so Electron child processes honor the configured heap cap.

## Why this ordering is safe

- Replay during `session/load` is historical transcript noise, not live turn output; dropping it at the ACP layer is intentional.
- Mode/config state still comes from the `session/load` response payload, so dropping replay notifications does not lose session setup data.
- Forking the consumer before config RPCs only changes startup ordering; it does not alter normal turn handling after the session is ready.

## Tests

- `AcpJsonRpcConnection.test.ts`: resumed `session/load` path
- `GrokAdapter.test.ts`: compaction detection + existing scoping helpers
- `backendNodeOptions.test.ts`: desktop heap arg wiring

### DIFF
--- a/apps/desktop/src/backendNodeOptions.test.ts
+++ b/apps/desktop/src/backendNodeOptions.test.ts
@@ -45,6 +45,7 @@ describe("withBackendHeapLimitArg", () => {
   it("adds a heap limit when none exists", () => {
     expect(withBackendHeapLimitArg("--trace-warnings", 4096)).toEqual([
       "--max-old-space-size=4096",
+      "--js-flags=--max-old-space-size=4096",
     ]);
   });
 
@@ -64,6 +65,6 @@ describe("resolveBackendNodeArgs", () => {
         existingNodeOptions: "--enable-source-maps",
         totalMemoryBytes: gb(24),
       }),
-    ).toEqual(["--max-old-space-size=6144"]);
+    ).toEqual(["--max-old-space-size=6144", "--js-flags=--max-old-space-size=6144"]);
   });
 });
--- a/apps/desktop/src/backendNodeOptions.test.ts
+++ b/apps/desktop/src/backendNodeOptions.test.ts
@@ -45,7 +45,6 @@ describe("withBackendHeapLimitArg", () => {
   it("adds a heap limit when none exists", () => {
     expect(withBackendHeapLimitArg("--trace-warnings", 4096)).toEqual([
       "--max-old-space-size=4096",
-      "--js-flags=--max-old-space-size=4096",
     ]);
   });
 
@@ -65,6 +64,6 @@ describe("resolveBackendNodeArgs", () => {
         existingNodeOptions: "--enable-source-maps",
         totalMemoryBytes: gb(24),
       }),
-    ).toEqual(["--max-old-space-size=6144", "--js-flags=--max-old-space-size=6144"]);
+    ).toEqual(["--max-old-space-size=6144"]);
   });
 });
--- a/apps/desktop/src/backendNodeOptions.ts
+++ b/apps/desktop/src/backendNodeOptions.ts
@@ -48,7 +48,10 @@ export function withBackendHeapLimitArg(
     return [];
   }
 
-  return [`--max-old-space-size=${maxOldSpaceMb}`];
+  return [
+    `--max-old-space-size=${maxOldSpaceMb}`,
+    `--js-flags=--max-old-space-size=${maxOldSpaceMb}`,
+  ];
 }
 
 export function resolveBackendNodeArgs(input: {
--- a/apps/desktop/src/backendNodeOptions.ts
+++ b/apps/desktop/src/backendNodeOptions.ts
@@ -48,10 +48,7 @@ export function withBackendHeapLimitArg(
     return [];
   }
 
-  return [
-    `--max-old-space-size=${maxOldSpaceMb}`,
-    `--js-flags=--max-old-space-size=${maxOldSpaceMb}`,
-  ];
+  return [`--max-old-space-size=${maxOldSpaceMb}`];
 }
 
 export function resolveBackendNodeArgs(input: {
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -3829,6 +3829,55 @@ describe("ProviderRuntimeIngestion", () => {
     expect(activity?.tone).toBe("info");
   });
 
+  it("projects context compaction completion and failure into thread activities", async () => {
+    const harness = await createHarness();
+    const now = new Date().toISOString();
+
+    harness.emit({
+      type: "item.completed",
+      eventId: asEventId("evt-thread-compaction-completed"),
+      provider: "grok",
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      payload: {
+        itemType: "context_compaction",
+        status: "completed",
+      },
+    });
+    harness.emit({
+      type: "item.completed",
+      eventId: asEventId("evt-thread-compaction-failed"),
+      provider: "grok",
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      payload: {
+        itemType: "context_compaction",
+        status: "failed",
+        detail: "Compaction was interrupted",
+      },
+    });
+
+    const thread = await waitForThread(harness.engine, (entry) =>
+      entry.activities.some(
+        (activity: ProviderRuntimeTestActivity) =>
+          activity.kind === "context-compaction" &&
+          activity.summary === "Context compaction failed",
+      ),
+    );
+
+    const completed = thread.activities.find(
+      (candidate: ProviderRuntimeTestActivity) =>
+        candidate.kind === "context-compaction" && candidate.summary === "Context compacted",
+    );
+    expect(completed?.tone).toBe("info");
+    const failed = thread.activities.find(
+      (candidate: ProviderRuntimeTestActivity) =>
+        candidate.kind === "context-compaction" &&
+        candidate.summary === "Context compaction failed",
+    );
+    expect(failed?.tone).toBe("error");
+  });
+
   it("projects Codex task lifecycle chunks into thread activities", async () => {
     const harness = await createHarness();
     const now = new Date().toISOString();
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
@@ -1104,6 +1104,29 @@ function runtimeEventToActivities(
     }
 
     case "item.completed": {
+      // Providers (Grok auto-compaction, Pi compaction_end) close their
+      // compaction rows via item.completed; without this branch the earlier
+      // "Compacting conversation..." activity never resolves.
+      if (event.payload.itemType === "context_compaction") {
+        const failed = event.payload.status === "failed";
+        return [
+          {
+            id: event.eventId,
+            createdAt: event.createdAt,
+            tone: failed ? "error" : "info",
+            kind: "context-compaction",
+            summary: failed ? "Context compaction failed" : "Context compacted",
+            payload: toActivityPayload({
+              itemType: event.payload.itemType,
+              status: event.payload.status,
+              ...(event.payload.detail ? { detail: truncateDetail(event.payload.detail) } : {}),
+              ...activityDataField(event.payload.data),
+            }),
+            turnId: toTurnId(event.turnId) ?? null,
+            ...maybeSequence,
+          },
+        ];
+      }
       if (!isToolLifecycleItemType(event.payload.itemType)) {
         return [];
       }
--- a/apps/server/src/provider/Layers/GrokAdapter.test.ts
+++ b/apps/server/src/provider/Layers/GrokAdapter.test.ts
@@ -7,6 +7,7 @@ import { TurnId } from "@t3tools/contracts";
 import { describe, expect, it } from "vitest";
 
 import {
+  isGrokContextCompactionToolCall,
   isRenderableGrokAssistantDelta,
   mergeGrokModelDescriptors,
   parseXaiLanguageModelDescriptors,
@@ -42,6 +43,27 @@ describe("GrokAdapter runtime event scoping", () => {
       toolCallId: "call-1",
       providerToolCallId: "call-1",
     });
+  });
+
+  it("detects Grok compaction tool calls for context compaction UI rows", () => {
+    expect(
+      isGrokContextCompactionToolCall({
+        toolCallId: "tool-1",
+        kind: "other",
+        status: "inProgress",
+        title: "Compacting conversation context",
+        data: {},
+      }),
+    ).toBe(true);
+    expect(
+      isGrokContextCompactionToolCall({
+        toolCallId: "tool-2",
+        kind: "execute",
+        status: "completed",
+        title: "Run tests",
+        data: {},
+      }),
+    ).toBe(false);
   });
 
   it("only treats visible assistant text as renderable Grok content", () => {
--- a/apps/server/src/provider/Layers/GrokAdapter.ts
+++ b/apps/server/src/provider/Layers/GrokAdapter.ts
@@ -937,7 +937,11 @@ export function makeGrokAdapter(
         const startedAt = Date.now();
         while (true) {
           const now = Date.now();
-          const lastActivityAt = ctx.lastTurnActivityAt ?? 0;
+          // Seed the quiet measurement from startedAt: a backlogged consumer
+          // may not have bumped lastTurnActivityAt yet, so always wait at
+          // least one full quiet interval after the prompt response before
+          // deciding the outcome.
+          const lastActivityAt = Math.max(ctx.lastTurnActivityAt ?? 0, startedAt);
           if (
             now - lastActivityAt >= GROK_COMPACT_OUTCOME_QUIET_MS ||
             now - startedAt >= GROK_COMPACT_OUTCOME_MAX_WAIT_MS
--- a/apps/server/src/provider/Layers/GrokAdapter.ts
+++ b/apps/server/src/provider/Layers/GrokAdapter.ts
@@ -126,6 +126,16 @@ const GROK_COMPACT_TIMEOUT_MS = GROK_TURN_IDLE_TIMEOUT_MS;
 // drop compaction-shaped tool updates) for this long so those events cannot
 // be attributed to the next active turn.
 const GROK_COMPACT_ABANDON_QUIET_MS = 5_000;
+// Bounded wait for the forked post-timeout cancel to be written before the
+// next prompt is dispatched. stdio delivers in order, so once the cancel is
+// on the wire it cannot cancel a prompt written after it; a fully wedged
+// child never confirms, hence the cap.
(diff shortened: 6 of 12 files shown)


Emanuele-web04:
Pushed fe443f7 to this branch after reviewing against #278. The core layered fix (dropping `session/update` in `AcpSessionRuntime` until `start()` completes + forking the Grok consumer before the config RPCs) is sound and kept as-is. What changed:

**Reverted the `--js-flags` addition.** The desktop backend child is spawned with `ELECTRON_RUN_AS_NODE=1`, i.e. a plain Node process. Verified against the repo's Electron binary: with the added flag the child exits immediately with `bad option: --js-flags=...` (code 9) and the backend would crash-loop; `--max-old-space-size` alone (already on main) boots fine and caps the heap in Node mode.

**Fixed two typecheck failures in `compactThread`** (`bun typecheck` was failing on this branch): `Cause.failureOption` doesn't exist in the pinned effect package — replaced with `Cause.squash` per repo convention — and the `thread.state.changed` payload contract is `{ state, detail? }`, not `{ state, reason }`.

**Hardened the new compaction flow:**
- The `compact|summariz` title heuristic now only applies between turns. Previously any live tool call whose title mentioned "compact"/"summarize" (e.g. a shell command) was hijacked into a compaction row and its real tool row swallowed. Resume replay also stays suppressed instead of leaking compaction-looking replay rows.
- `sendTurn` now rejects while compaction is in flight. It doesn't take the thread lock, so a mid-compaction message would have raced a second `session/prompt` whose events the compaction suppression silently drops (turn stuck on "Working").
- Since `applyRequestedSessionConfiguration` now runs after the session is registered (scope transferred), a config failure tears the session down instead of leaking a live child the start-scope finalizer no longer owns.
- Interruption during the `/compact` prompt propagates instead of being reported as a failed compaction.

`bun fmt` / `bun lint` / `bun typecheck` and the touched test files all pass locally.

Emanuele-web04:
@codex review

Emanuele-web04:
Addressed all 6 review findings in af931fb10 — each one was verified against the head and made sense:

- **P1 replay gate** (`AcpSessionRuntime.ts`): the session/update gate now stays closed after a successful `session/load` until the adapter attaches its consumer — `getEvents()` opens it at attach time, so a post-load replay can no longer accumulate in the unbounded queue. Fresh sessions keep the buffer-from-start behavior. (Note: this also applies to CursorAdapter resumes, where replay arriving during its config RPCs is now dropped instead of buffered — aligned with the replay-suppression intent.)
- **Cancelled /compact** (`GrokAdapter.ts`): a successful `PromptResponse` with `stopReason: "cancelled"` now emits a failed compaction row and fails the RPC instead of persisting `compacted`.
- **Compaction vs in-flight sendTurn**: new `turnStarting` flag set in the same synchronous block as sendTurn's compaction check; `compactThread` rejects while a turn is active *or* starting, closing the pre-`activeTurnId` gap.
- **Compaction completion ingestion** (`ProviderRuntimeIngestion.ts`): `item.completed` with `itemType: "context_compaction"` now projects a resolving activity ("Context compacted" / "Context compaction failed") — this also fixes Pi's `compaction_end`, which was dropped the same way.
- **Startup interruption cleanup**: after scope ownership transfer, the remaining startup steps run under `Effect.onExit`, so interruption (not just typed errors) tears down the registered session, notification fiber, and child.
- **Fixed replay timeout**: suppression no longer ends at 1.5s — it stays active until the replay is quiet (30s hard backstop with warning). Startup still unblocks after 1.5s, but `sendTurn`/`compactThread` stay gated on the deferred until the replay actually settles, so late historical deltas can't corrupt the first post-resume turn.

@codex review

Emanuele-web04:
Addressed all 5 second-round findings in ecd0b8d8f:

- **Cancellable replay-gated turns**: `interruptTurn` now flags a turn that is still starting (no prompt fiber yet), and `startGrokTurn` re-checks interrupted/stopped state after the pre-prompt waits, right before assigning `activeTurnId` — a cancelled turn is never dispatched once the replay quiets.
- **Thread lock vs replay wait**: `compactThread` awaits `resumeReplayReady` *before* taking the thread lock, so stop/restart (which resolves the deferred) is never stalled behind the wait. If the session was restarted while waiting and a new replay window is settling, it rejects instead of blocking the lock.
- **Fresh-session updates during `session/new`**: the gate now opens *before* `createSession` on both fresh paths (plain new and failed-load fallback), restoring the buffering of early agent output; only a successful `session/load` keeps the gate closed until the consumer attaches.
- **`compactingThread` lifetime**: the inline clear after the `/compact` response is gone; the `ensuring` block owns it, so `sendTurn` keeps rejecting until the completion/thread-state events have published.
- **Duplicate compaction rows**: during manual `/compact`, tool-call updates are emitted as progress only (`item.updated`/`inProgress`) — `compactThread` owns the single terminal row. Grok-initiated auto-compaction keeps emitting its own terminal row since it has no other completion source.

Full server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 3 third-round findings in 4f66f4a28:

- **P1 — lock held during compaction**: the thread lock now only guards claiming the compaction slot (`claimGrokCompactionSlot`); the `/compact` prompt runs outside it, so stop/restart can always reach `stopSessionInternal` and kill a hung child. Since lock serialization no longer covers the prompt, a concurrent `/compact` is rejected while one is in flight.
- **Duplicate terminal rows**: successful manual compaction now emits only `thread.state.changed` (projected as "Context compacted manually"), matching OpenCode; the redundant `item.completed` row is gone. Failure/cancellation keep their `item.completed` failed row (no `thread.state.changed` fires on those paths).
- **Interrupt before fiber registration**: the prompt fiber re-checks `pendingTurnInterrupted`/`stopped` via `Effect.suspend` immediately before dispatching `session/prompt` and self-interrupts through the existing cancelled-turn path, closing the window between the pre-dispatch check and `activePromptFiber` assignment.

Full server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 3 fourth-round findings in fc2dc3a9b:

- **Pre-start interrupts as cancellation**: the pre-dispatch failure check is gone. Interrupts that land during the pre-prompt waits now flow through the prompt fiber's dispatch guard, which self-interrupts into the existing cancelled-turn path — the turn completes `turn.started` → `turn.completed(cancelled)` instead of failing `sendTurn` and having the reactor mark the session errored.
- **Failed compaction tool terminals**: while `compactingThread` is set, a `failed` compaction tool call records `compactionFailedToolDetail` (via `readAcpFailedToolDetail`, mirroring how normal turns track failed tools). `runGrokCompaction` checks it after a successful non-cancelled prompt and emits the failed row + fails the RPC instead of persisting `compacted`.
- **Stale progress rows**: the web work log now collapses a `context-compaction` progress row ("Compacting conversation...") into the terminal row that follows it, for all providers that emit both lifecycles (Grok auto-compaction, Pi). Terminal rows never swallow a subsequent compaction's progress row.

Web suite (2276 tests), server suite (1520), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 3 fifth-round findings in 4ad44250b:

- **P1 — pre-start cancel vs post-return running write**: `updateSessionBindingFromRuntimeEvent` now records the terminal turn id (in-memory, per thread) *before* that event's own binding upsert, and `ProviderService.sendTurn` checks it immediately before its post-dispatch "running" upsert. If the terminal event's upsert can precede sendTurn's, the record is by construction already visible to the check, so a settled turn is never re-marked running with a stale active turn. (The guard is general — it also covers any ultra-fast turn completion racing that write, not just pre-start cancels.)
- **Fresh-session merge state**: the tool-call and assistant-segment refs are no longer reset after `session/new` on fresh paths — updates accepted while the request was in flight keep their merge/segment state, so continuations after the response still merge and `prompt()` can close open segments. Only the resumed replay-dropping path starts clean.
- **Hung /compact**: the compaction prompt is bounded by a hard timeout (the turn idle timeout value). On expiry: best-effort `session/cancel`, a failed compaction row, and an RPC failure — `compactingThread` unblocks instead of wedging the thread.

Server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 3 sixth-round findings in eff561f82:

- **Abandoned compaction stragglers**: a timed-out `/compact` now opens a short quiet window (`compactionQuietUntil`, 5s past the best-effort cancel). New turns wait it out before dispatching — so stale updates arrive while no turn is active and fall into the existing orphan suppression — and compaction-shaped tool updates are dropped outright during the window. `compactingThread` stays a pure send-blocking flag. (The window is only set on timeout; on a cancelled/failed prompt the response already arrived, meaning the child finished emitting for that prompt.)
- **Model selection persistence**: when the settled-turn guard skips the "running" upsert, `sendTurn` still writes the durable metadata — `modelSelection` merges into the runtime payload and the resume cursor is stored, while `status`/`activeTurnId` stay untouched (the directory upsert keeps existing status when omitted and merges payload per key, so no read-modify-write race is introduced).
- **Concurrent turn-start**: `sendTurn` now rejects while `turnStarting` is already set, in the same synchronous block as the compaction check — a second call can no longer clear the first turn's `pendingTurnInterrupted` flag or race two ACP prompts.

Server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed both seventh-round findings in a6ad9fa12:

- **P1 — awaited best-effort cancel**: the post-timeout `session/cancel` is now forked into the session scope instead of awaited — a silent child can no longer hold `runGrokCompaction` hostage before the failure event and the `ensuring` that clears `compactingThread`. The same fix is applied to the turn idle watchdog's cancel (`failGrokTurnAsTimedOut`), which had the same hazard: a hung cancel there would have blocked the prompt-fiber interrupt.
- **Settled-turn guard atomicity**: binding writes are now serialized per thread (`Semaphore.makeUnsafe(1)` per `threadId`, synchronous creation so no double-mint). Both the runtime-event handler's record+upsert and `sendTurn`'s check+upsert run under the lock, so a terminal event can no longer complete between the map check and the "running" write. Lifecycle events are low-frequency, so the lock adds no meaningful contention (content deltas don't take this path).

Server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 4 eighth-round findings in 2865e2a52:

- **P1 — undrained raw notification queue**: `AcpSessionRuntime` now forks a lifetime drain of `acp.raw.notifications`. The protocol was offering every notification into that unbounded queue in addition to the `handleSessionUpdate` callback path, so the resume replay could still accumulate there regardless of the accepting gate — this was the real remaining leg of the original OOM. Great catch. (Callback delivery is unaffected; nothing else consumes that stream.)
- **Post-timeout cancel ordering**: the forked cancel fiber is stored on the session context, and both the next `sendTurn` and the next `/compact` wait for it (bounded 10s) before dispatching. Since stdio delivers in order, a cancel that is on the wire before the new prompt cannot cancel it; a fully wedged child hits the cap and falls to the idle watchdog.
- **Queued compaction failures**: `runGrokCompaction` now waits for inbound consumer activity to go quiet (200ms quiet / 2s cap) after the prompt resolves before reading `compactionFailedToolDetail`, so a failed tool update still sitting in the event queue is seen before success is declared.
- **Same-millisecond row ordering**: `compareActivitiesByOrder` ranks the `context-compaction` progress row before terminal rows at equal `createdAt`, so the work-log collapse always folds the pair regardless of random event ids.

Server suite (1520 tests), web suite (2277), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed the ninth-round finding in the latest push: `settleGrokCompactionOutcome` now seeds its quiet measurement from the call time (`Math.max(lastTurnActivityAt, startedAt)`), so every manual compaction waits at least one full 200ms quiet interval after the prompt response before the outcome is decided — a backlogged consumer with a stale activity timestamp can no longer short-circuit the wait.

Server suite (1520 tests), fmt, lint, and typecheck all pass.

@codex review

Emanuele-web04:
Addressed all 5 tenth-round findings in 1e39eb216:

- **P2 — settled live-fallback turns persisted as running** (both `ProviderService.ts` findings): the settled-turn branch is the first directory write for a thread on the live-fallback path, and `upsert` defaults a new row's omitted status to `running`. It now checks for an existing binding under the binding-write lock and writes an explicit `status: "stopped"` when it creates the first row; existing rows keep the status the runtime-event handler already wrote. Covered by a new regression test (`persists the first binding row as stopped when the turn settles pre-write`).
- **P2 — late turn tool calls misfiled as auto-compaction**: the adapter now records each live turn's provider tool-call ids (`ctx.turnToolCallIds`, cleared when the next turn dispatches). A queued `ToolCallUpdated` processed after the prompt response cleared `activeTurnId` is emitted for its originating turn (same scoped item id, so the row resolves in place) instead of falling into the between-turn compaction heuristic — even when its title mentions "compact"/"summarize".
- **P2 — sessions exposed before config settles**: every session now carries a `sessionConfigReady` deferred, resolved after `applyRequestedSessionConfiguration` completes in startup (and by `stopSessionInternal`, mirroring `resumeReplayReady`, so failed startups never strand waiters). `startGrokTurn` and `compactThread` await it before prompting, so a turn routed in via `hasSession` mid-startup can no longer run with provider defaults.
- **P2 — compaction after stop/restart targets the new session**: `claimGrokCompactionSlot` now takes the pre-lock context and fails with a validation error when the locked session context is a different instance (a stop/restart won the lock while `compactThread` was waiting on the replay deferred), instead of sending `/compact` to the freshly restarted session.

`bun fmt`, `bun lint`, `bun typecheck`, and the provider test suites all pass.

@codex review

Emanuele-web04:
Addressed all 3 eleventh-round findings in 857022494:

- **P2 — quiet window expiring during the cancel wait**: `waitForAbandonedGrokCompaction` now restarts `compactionQuietUntil` (`max(current, now + GROK_COMPACT_ABANDON_QUIET_MS)`) after the bounded cancel-fiber join, so stragglers arriving just after a slow `session/cancel` drains are still held off and dropped instead of being attributed to the next turn.
- **P2 — late tool ids with no map entry**: rather than trying to capture ids earlier (a single-update tool never gives the consumer a chance), the turn now stays active until the queued backlog drains: `AcpSessionRuntime` exposes `pendingSessionUpdateCount` (parsed event-queue depth; raw notifications dispatch synchronously into it), and the prompt fiber's settle paths wait — bounded at 1s, returning immediately when the queue is already empty — before snapshotting `activeTurnFailedToolDetail`/content flags and clearing `activeTurnId`. Queued live-turn updates therefore process inside the active-turn window with normal attribution; the `turnToolCallIds` map remains as a second line of defense past the cap.
- **P2 — dead-startup sends after the config gate opens**: `startGrokTurn` re-checks `ctx.stopped` after the `sessionConfigReady`/`resumeReplayReady` awaits (both are resolved by `stopSessionInternal` to unblock waiters) and fails with `ProviderAdapterSessionNotFoundError` instead of emitting turn lifecycle events for a session that was torn down mid-startup. The compaction path is already covered: `claimGrokCompactionSlot` goes through `requireSession`, which rejects stopped contexts.

`bun fmt`, `bun lint`, `bun typecheck`, and the provider + ACP test suites all pass.

@codex review

Emanuele-web04:
Addressed all 4 twelfth-round findings in de3834156:

- **P2 — in-flight items invisible to the drain**: dropped the `Queue.size` probe entirely. `AcpSessionRuntime` now counts every parsed event it offers (`sessionUpdatesEnqueuedCount`, all offers routed through one counting `offer` fn), and the Grok notification consumer bumps `ctx.sessionUpdatesProcessed` only after a handler fully runs (`Effect.ensuring`). `waitForGrokQueuedTurnEventsDrained` snapshots the enqueued count and waits until processed catches up — correct under stream chunk buffering and mid-flight handlers, still zero-cost when the consumer kept up.
- **P2 — compaction success declared over a backlogged failure**: `settleGrokCompactionOutcome` now runs that same drain first, so a failed compaction `ToolCallUpdated` already in the queue lands in `compactionFailedToolDetail` before the outcome is decided; the 200ms quiet window afterwards only covers in-transit stragglers.
- **P2 — stop racing the pre-prompt work**: `startGrokTurn` re-checks `ctx.stopped` immediately before assigning `activeTurnId`/publishing `turn.started`, so a stop that lands during config RPCs or attachment reads fails the send as `ProviderAdapterSessionNotFoundError` instead of emitting a phantom cancelled turn after `session.exited`.
- **P2 — single completion slot per thread**: `ProviderService` now keeps a small per-thread FIFO (cap 8) of recently completed turn ids; each `sendTurn` consumes its own marker under the binding-write lock, so overlapping pre-write settlements can no longer evict each other and re-mark the thread as running with a stale `activeTurnId`.

`bun fmt`, `bun lint`, `bun typecheck`, and the provider + ACP test suites all pass.

@codex review

PR #198, opened 2026-06-14
[codex] Add bundled Wandy runtime

reposlayer:
## Summary
- rename the bundled desktop automation runtime from Do The Thing/dothething/DTT to Wandy across package names, CLI commands, settings, prompts, scripts, and desktop wiring
- move the workspace package to `packages/wandy` with `@t3tools/wandy`, `wandy`, and `wandy-mcp` entrypoints so release smoke can resolve the workspace dependency
- rebuild the bundled macOS Wandy app with the `com.synara.wandy` bundle identifier and refreshed code signature metadata

## Validation
- `bun run fmt`
- `bun run lint` (passes with existing repo warnings, 0 errors)
- `bun run typecheck`
- `bun run test` (128 files passed, 1 skipped; 1238 tests passed, 6 skipped)
- `node scripts/release-smoke.ts`
- `bun run build:macos` from `packages/wandy`
- `codesign --verify --deep --strict --verbose=2 packages/wandy/dist/Wandy.app`
- `WANDY_SUPPRESS_ONBOARDING=1 packages/wandy/bin/wandy doctor --json`
- `swift build --product Wandy`, `swift build --product WandySmokeSuite`, `swift build --product WandyFixture`
- tracked-code and bundled-runtime string scans for old Do The Thing/dothething package names

## Notes
- `swift test` is blocked in this local Command Line Tools environment because XCTest is unavailable (`no such module 'XCTest'`). The Swift products above build successfully.
- Go package tests were not run locally because `go` is not installed on this machine.

PR #286, opened 2026-07-02
[codex] Add bundled Wandy runtime (updated: conflicts resolved + review fixes)

Emanuele-web04:
Supersedes #198, which was detached from its head branch and can no longer receive updates. This PR carries the original Wandy runtime work, freshly merged with `main` (v0.3.7, was 189 commits behind), with all conflicts resolved plus a full logic-review and performance pass.

## Original scope (from #198)
- Rename the bundled desktop automation runtime from Do The Thing/dothething/DTT to Wandy across package names, CLI commands, settings, prompts, scripts, and desktop wiring
- Move the workspace package to `packages/wandy` with `@t3tools/wandy`, `wandy`, and `wandy-mcp` entrypoints
- Rebuild the bundled macOS Wandy app with the `com.synara.wandy` bundle identifier

## Merge with main
All 11 conflicts resolved; notable resolutions:
- Cursor/Grok/Gemini adapters: main's `appendFileAttachmentsPromptBlock` refactor kept, Wandy prompt context now wraps the combined prompt text
- Settings: `enableWandy` kept alongside main's `enableAssistantStreaming: true` default and `enableProviderUpdateChecks`
- toolCallLabel: Wandy alias labeling kept; main's structured-payload test repointed at a neutral server

## Review fixes
- **Wire the "Enable Wandy" setting to the actual gate** — it persisted but was never read; `SYNARA_ENABLE_WANDY` was hardcoded `"1"`. The server now mirrors the setting into the env on load and update (boot-env snapshot so the operator env escape hatch still wins). Takes effect for new sessions without a restart.
- **Fix fatal startup crash path** — stable-helper fingerprinting ran outside its error boundary; a transient FS error would quit Synara instead of falling back to the bundled launcher.
- **Fix Gemini rollback amnesia** — the Wandy resume-skip gate also discarded rollback's cloned session snapshot, silently resetting the model to an empty conversation. Skip now applies only to best-effort reconnects.
- **Scope Wandy process cleanup** to this install's app bundles (previously killed every Wandy process system-wide, including other Synara instances).
- `SYNARA_ENABLE_WANDY=0` now skips stable-helper install/fingerprint work entirely.
- Codex/Claude prompts only include Wandy routing instructions when the wandy MCP server is actually registered.
- Stale wandy/legacy Codex config sections are cleaned even when no launcher resolves.
- Windows: forward-slash launcher paths now upgrade to the native runtime; bin launcher resolution checks executability.

## Performance
- Stable-helper staleness check: full-content SHA-256 of both app bundles on the Electron main thread every launch → metadata fingerprint of the source bundle recorded in a manifest at install time (measured: install 5ms, reuse 0.4ms against the real bundle).
- Wandy launcher resolution (fs-stat chain on every turn/session) memoized with a 5s TTL; enable/disable gate stays uncached so toggles apply immediately.
- Desktop artifacts stage only their own platform's Wandy runtime (Windows/Linux installers no longer embed the macOS .app and each other's binaries).

## Validation
- `bun fmt`, `bun lint` (0 errors), `bun typecheck`
- `bun run test`: 10/10 turbo tasks; server 1533 passed / 6 skipped, web 2274 passed, plus shared/desktop/contracts/scripts
- Suites now pin `SYNARA_ENABLE_WANDY=0` in the base vitest config so results are deterministic under desktop-app shells that export `DPCODE_MODE=desktop`
- Runtime spot-check: `wandy doctor --json` on the bundled binary; stable-helper install/reuse exercised against the real signed `Wandy.app`


Emanuele-web04:
All 6 review findings verified and fixed in 4fe840840:

1. **run_sequence off macOS** — confirmed absent from the Linux/Windows runtimes; the tool catalog and prompt instructions are now platform-aware (`wandyMcpToolNamesForPlatform`), so agents are never told about a tool that doesn't exist.
2. **Fallback launcher discarded** — `resolveWandyLauncherPath` now uses the bundled launcher when the stable-helper install fails, instead of disabling Wandy MCP.
3. **wandy-mcp default mode** — the bin entry now defaults to `mcp` when invoked without arguments (fixed in both the checked-in shim and the npm package generator). Verified: bare invocation starts the stdio server instead of printing help.
4. **Installer legacy self-delete** — reruns no longer delete the entry they just wrote when the legacy alias equals the target name; fixed in all four installers (Claude/Gemini/opencode/Codex) in both helper copies. Verified with a rerun-idempotency script.
5. **Codex home overlay mismatch** — added a shared `shouldUseCodexHomeOverlay` predicate used by both `buildCodexProcessEnv` and `resolveActiveCodexHomeWritePath`, so image-path prediction always follows the actual overlay decision (including the Wandy-only case).
6. **Platform-dependent tests** — the Wandy path assertions now pin `platform: "darwin", arch: "arm64"` explicitly, so the shared suite passes on non-macOS runners.

Emanuele-web04:
@codex review

Emanuele-web04:
@codex-bot please review/check the latest head.

@codex review

Emanuele-web04:
@codex-bot please review/check the latest head.

@codex review

PR #294, opened 2026-07-03
Add Hermes provider integration

sangnom:
## Summary
- Add Hermes as a first-class Synara provider across contracts, shared models, server runtime wiring, provider health, and web provider/model selection UI.
- Add Hermes adapter, service registration, provider registry/runtime integration, settings support for `hermesBinaryPath`, and provider health probing with a Hermes-specific timeout.
- Add/extend tests for Hermes adapter, provider health/status, provider registry, app settings, orchestration contracts, provider/model picker flow, composer/session logic, and related UI state.

## Verification
- `bun run fmt:check`
- `bun run lint` — exits 0; existing warnings remain, no errors
- `bun run typecheck`
- `bun run --filter=t3 typecheck`
- `bun run test src/provider/Layers/ProviderHealth.test.ts` — 82 tests passed
- `bun run test src/orchestration/Layers/ProviderCommandReactor.test.ts` — 56 tests passed
- Live smoke: Hermes provider health cache reached `ready`, `available: true`, version `0.16.0`; Hermes appeared enabled in provider picker.

## Notes
- Hermes CLI startup can take ~9–16 seconds in this environment, so Hermes health uses a longer provider-specific timeout to avoid false unavailable states.
- Dev smoke required `~/.local/bin` in the server PATH so the `hermes` binary can be discovered.


sangnom:
Closing this PR because the branch was pushed from a local copy and should not remain public.

PR #298, opened 2026-07-03
fix(acp): prevent assistant message ID collisions across runtime restarts

kartikkabadi:
## Summary

- Assistant message IDs were built as `assistant:{sessionId}:segment:{N}`, where the segment index resets to 0 on every `AcpSessionRuntime` creation (server restart or session resume).
- When a session was resumed with the same `sessionId`, post-restart assistant segments collided with pre-restart ones. The projection's `ON CONFLICT DO UPDATE` then overwrote earlier messages in the middle of the transcript instead of appending new ones, making assistant replies silently disappear.
- Add a per-runtime-instance UUID (8 chars) to the ID so segments from different runtime instances never collide, even when the `sessionId` and segment index are identical.

Affects **Cursor** when the provider does not supply stable `itemId`s and the runtime uses the fallback ID path. Grok already scopes fallback item IDs by turn via `scopeGrokRuntimeItemIdForTurn` (`grok:{turnId}:{itemId}`), and turn IDs are unique across restarts, so Grok was not susceptible to this collision. Non-ACP providers (Codex, Claude, Gemini, OpenCode, Pi) use their own ID generation and are unaffected.

**Known behavior change:** if the server restarts mid-stream, post-restart continuation appears as a new message rather than overwriting the partial one. That is strictly better than silent data loss.

#### Test plan
- [x] Unit test: `assistantItemId` format contract (distinct IDs for different runtime instance IDs)
- [x] Integration test: two separate `AcpSessionRuntime` spawns with the same resumed session produce distinct fallback assistant item IDs
- [x] All existing tests pass
- [x] `bun fmt`, `bun lint`, `bun typecheck` all pass

Generated with [Devin](https://devin.ai)

### DIFF
--- a/apps/server/src/provider/acp/AcpJsonRpcConnection.test.ts
+++ b/apps/server/src/provider/acp/AcpJsonRpcConnection.test.ts
@@ -103,6 +103,46 @@ describe("AcpSessionRuntime", () => {
     ),
   );
 
+  it.effect(
+    "assigns distinct fallback assistant item ids across separate runtime instances",
+    () => {
+      const runtimeLayer = AcpSessionRuntime.layer({
+        spawn: {
+          command: bunExe,
+          args: [mockAgentPath],
+        },
+        cwd: process.cwd(),
+        resumeSessionId: "mock-session-1",
+        clientInfo: { name: "t3-test", version: "0.0.0" },
+        authMethodId: "test",
+      });
+
+      const collectFallbackAssistantItemId = Effect.gen(function* () {
+        const runtime = yield* AcpSessionRuntime;
+        yield* runtime.start();
+        const eventsFiber = yield* Stream.runCollect(Stream.take(runtime.getEvents(), 4)).pipe(
+          Effect.forkChild,
+        );
+        yield* runtime.prompt({
+          prompt: [{ type: "text", text: "hi" }],
+        });
+        const notes = Array.from(yield* Fiber.join(eventsFiber));
+        const delta = notes.find((note) => note._tag === "ContentDelta");
+        expect(delta?._tag).toBe("ContentDelta");
+        return delta?._tag === "ContentDelta" ? delta.itemId : undefined;
+      }).pipe(Effect.provide(runtimeLayer), Effect.scoped, Effect.provide(NodeServices.layer));
+
+      return Effect.gen(function* () {
+        const firstItemId = yield* collectFallbackAssistantItemId;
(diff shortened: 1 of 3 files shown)


Emanuele-web04:
@kartikkabadi Reviewed this — the bug is real and the fix is correct. I traced the full flow: the fallback ID `assistant:{sessionId}:segment:{N}` uses a per-runtime-instance segment counter (explicitly reset to 0 on the resumed-session path in `AcpSessionRuntime`), `CursorAdapter` passes `event.itemId` through unchanged, ingestion builds `assistant:{itemId}` message IDs, and the projection upserts with `ON CONFLICT (message_id) DO UPDATE SET text = excluded.text` — so post-resume segments really did overwrite earlier transcript messages. Nothing in the repo parses the old ID format, so changing it is safe.

A few minor, non-blocking notes:

- **Grok impact is overstated in the description.** `GrokAdapter` already scopes item IDs by turn via `scopeGrokRuntimeItemIdForTurn` (`grok:{turnId}:{itemId}`), and turn IDs are unique across restarts, so Grok wasn't actually susceptible. In practice this fixes Cursor only — worth a one-line correction in the PR body.
- **The test doesn't exercise the actual fix.** It calls `assistantItemId` with two hardcoded instance IDs, so it can't catch a regression like hoisting `runtimeInstanceId` to module scope (which would silently reintroduce the collision). A stronger test would spin up two runtime instances, but that needs a child-process harness, so acceptable as-is with the caveat noted.
- **Entropy is fine.** `randomUUID().slice(0, 8)` gives 32 random bits, and a collision would additionally require the same `sessionId` and overlapping segment indices — non-concern.

One residual limitation worth knowing (not a flaw in this PR): if the server restarts mid-stream, the continuation now lands in a new message rather than overwriting the old one. That's strictly better than silent data loss, but the pre-restart partial text will show as a separate completed message.

LGTM.

kartikkabadi:
Addressed the review notes:

- **PR description** — corrected to Cursor-only; Grok is already turn-scoped via `scopeGrokRuntimeItemIdForTurn`.
- **Test coverage** — added an integration test in `AcpJsonRpcConnection.test.ts` that spins up two separate `AcpSessionRuntime` instances (same resumed `sessionId`) and asserts their fallback assistant item IDs differ. The unit test remains as a format contract with a comment pointing to the integration test.
- **Mid-stream restart** — documented in the PR body as expected behavior (new message vs silent overwrite).

Thanks for the thorough trace.

PR #299, opened 2026-07-03
Add composer prompt history navigation

asssslay:
## What Changed

Added prompt history navigation to the chat composer.

- `ArrowUp` navigates to the most recent submitted prompt.
- Repeated `ArrowUp` moves backward through prompt history.
- `ArrowDown` moves forward through history.
- When history navigation reaches the newest edge, the original unsent draft is restored.
- History is derived from retained native user messages for the current chat, capped at 100 entries.
- Multiline prompts preserve normal editor behavior: history navigation starts at line boundaries, and recalled multiline prompts can continue navigating in both directions.

## Why

This makes it easier to reuse or edit recently submitted prompts without copying them from the transcript.

The implementation keeps scope tight by using the existing composer key handling path and deriving history from already-retained chat messages instead of adding new persisted storage. This avoids extra storage growth while preserving predictable per-chat behavior.

## UI Changes

Interaction-only change. No visual UI changes.

https://github.com/user-attachments/assets/412efeeb-8abb-40d3-b308-946c57f57b22

## Checklist

- [ ] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes


Emanuele-web04:
sheesh gonna add this soon

Emanuele-web04:
Hey @asssslay, thanks for the PR — really cool feature, and the pure logic extraction into `ChatView.logic.ts` with solid test coverage is appreciated. CI is green. A few findings from review, roughly in order of importance:

**1. Moving the caret inside a recalled prompt silently destroys the saved draft.**
The wiring clears `promptHistoryNavigationRef` on any composer change that isn't the expected recalled prompt, and Lexical's `OnChangePlugin` also fires on cursor-only movement. So: recall a prompt → ArrowLeft (change matches `expectedPromptHistoryPromptRef`, expected cleared, nav state kept) → ArrowLeft again (expected is now null, `applyingPromptHistoryNavigationRef` is false, nav state wiped). At that point the original unsent draft — which only lived in `state.draft` — is gone, with no way to get it back via ArrowDown. It's also inconsistent that the first cursor move keeps history mode alive while the second exits it. Suggestion: either keep history mode alive across cursor-only changes (ignore value-identical changes while the value still equals `history[state.index]`), like shells do, or exit on the *first* movement but restore the draft.

**2. ArrowDown on the first line of a recalled multiline prompt navigates history instead of moving into line 2.**
`cursorCanNavigateNewer` accepts first-line *or* last-line positions, and the "can navigate newer immediately after recalling a multiline prompt" test locks this in — so it's clearly intentional to make ArrowUp/ArrowDown symmetric right after recall. The cost is you can't arrow down into the body of a multiline prompt you just recalled until you click or edit. Defensible tradeoff, just want to confirm it's the UX you want.

**3. Stale index can swallow ArrowDown.**
In `resolvePromptHistoryNavigation`, the stale check uses `input.history[input.state.index] ?? input.currentPrompt`, so if history ever shrinks below `state.index` the check passes, and the "newer" branch can return `handled: true` with an unchanged prompt — the keypress gets eaten. Rare in practice, but comparing against `undefined` directly (treating out-of-range as stale) is strictly safer.

**4. History nav fires while an empty trigger menu is open.**
The menu handlers only claim ArrowUp/Down when `currentItems.length > 0`, so with an active trigger and zero matches (e.g. `@zzz`) ArrowUp replaces the whole prompt with a history entry while the picker UI is still up. Recoverable via ArrowDown, but a `!menuIsActive` guard on the history block would be cleaner.

**Nits**
- Adjacent duplicate prompts aren't deduped (asserted on purpose in tests), so ArrowUp between two identical sends looks like a no-op. Shells dedupe consecutive duplicates; a one-line skip in `derivePromptHistoryFromMessages` would feel better.
- The wiring adds three refs plus a `requestAnimationFrame` handshake to an already very large `ChatView.tsx`, and that ref dance is exactly what made finding 1 possible. Extracting a small `usePromptHistoryNavigation` hook would contain the state machine nicely.

The `selectionCollapsed` plumbing through `ComposerPromptEditor` and the kanban editor looks correct — real selection state is only read where it matters, and fallbacks default to `true`.

PR #292, opened 2026-07-02
Preserve archived profile stats after thread purge

Emanuele-web04:
## Summary
- Add a profile-stats archive so purged threads can be hard-deleted without losing lifetime metrics.
- Update profile stats queries to merge live projection data with archived rows for prompts, tokens, turns, skills, and most-worked project rankings.
- Add startup cleanup logic for old soft-deleted threads and migration support for the new archive tables.
- Expand coverage for archive/purge flows, retention-hidden threads, and deleted-project behavior.

## Testing
- Not run (PR content only).
- Covered by new tests in `apps/server/src/profileStatsArchive.test.ts` and updates to `apps/server/src/profileStats.test.ts`.
- Added migration coverage through `apps/server/src/persistence/Migrations/050_ProfileStatsArchive.ts`.

PR #301, opened 2026-07-04
Fix Git writing model picker for OpenCode and Kilo discovery

### LINKED ISSUE #289: [Bug]: OpenCode only shows GPT-5 when choosing Git writing model
### Before submitting

### Area

apps/web

### Steps to reproduce

Image

### Expected behavior

it should actually load the models available in opencode

### Actual behavior

it seems to just load the placeholder GPT-5 model

### Impact

Minor bug or occasional failure

### Version or commit

_No response_

### Environment

_No response_

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

kartikkabadi:
## Summary
- Wire runtime-discovered OpenCode and Kilo models into the Settings → Models **Git writing model** dropdown (same discovery path the composer already uses).
- Pass the selected git-writing `textGenerationModelSelection` through git actions so commit/PR generation routes to OpenCode/Kilo instead of misrouting via slug heuristics and surfacing **Action failed**.
- Fix CI blockers: `exactOptionalPropertyTypes` on git text-generation params, diff-summary React Query cache keys for model selection, and value import for `DEFAULT_GIT_TEXT_GENERATION_MODEL`.

## Context
Fixes #289. The reporter confirmed composer model discovery works; only the Git writing picker was stale. Their follow-up showed the custom-slug workaround still failed because git actions did not carry provider selection.

## Test plan
- [x] Open Settings → Models on a machine with OpenCode configured; wait a moment and confirm the Git writing model dropdown lists live OpenCode models (not just `OpenAI GPT-5`).
- [x] Select an OpenCode git-writing model and run Commit (with AI message) — should succeed instead of **Action failed**.
- [x] Repeat with a Kilo git-writing model.
- [x] `bun run fmt:check`, `bun run typecheck`, and targeted vitest on `apps/web/src/appSettings.test.ts`, `packages/contracts/src/git.test.ts`, `apps/server/src/git/Layers/ProviderTextGeneration.test.ts`, `apps/server/src/git/Layers/GitManager.test.ts`

## Status
Rebased on latest `upstream/main` (includes #297, #298, #292). Single squashed commit; CI pending.

## Demo
Demo recording:

https://github.com/user-attachments/assets/3d4abe70-f3ca-4918-9f89-b1862dfdf14d

### DIFF
--- a/apps/server/src/git/Layers/GitManager.ts
+++ b/apps/server/src/git/Layers/GitManager.ts
@@ -6,6 +6,7 @@ import type {
   GitActionProgressEvent,
   GitActionProgressPhase,
   GitStackedAction,
+  ModelSelection,
   ProviderStartOptions,
 } from "@t3tools/contracts";
 import {
@@ -25,6 +26,7 @@ import {
 import { GitCore } from "../Services/GitCore.ts";
 import { GitHubCli, type GitHubPullRequestSummary } from "../Services/GitHubCli.ts";
 import { TextGeneration } from "../Services/TextGeneration.ts";
+import { buildGitTextGenerationCallInput } from "../textGenerationSelection.ts";
 import { ServerConfig } from "../../config.ts";
 
 const COMMIT_TIMEOUT_MS = 10 * 60_000;
@@ -73,6 +75,13 @@ interface BranchHeadContext {
   headRepositoryNameWithOwner: string | null;
   headRepositoryOwnerLogin: string | null;
   isCrossRepository: boolean;
+}
+
+interface GitTextGenerationParams {
+  textGenerationModel?: string | undefined;
+  textGenerationModelSelection?: ModelSelection | undefined;
+  codexHomePath?: string | undefined;
+  providerOptions?: ProviderStartOptions | undefined;
 }
 
 interface FailedLocalHandoffRecovery {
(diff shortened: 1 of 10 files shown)


Emanuele-web04:
@kartikkabadi I did a code-wise pass on this PR. The core fix direction looks right: the picker now uses discovered Codex/Kilo/OpenCode model options, and git actions carry a provider-aware `textGenerationModelSelection` through to the server, so the OpenCode/Kilo routing issue should be addressed.

A few things need fixing before this is merge-ready:

- **Blocking:** `apps/web/src/components/GitActionsControl.tsx:6` imports `DEFAULT_GIT_TEXT_GENERATION_MODEL` via `import type`, but it is used as a runtime value at `apps/web/src/components/GitActionsControl.tsx:323`. This should be a normal value import, otherwise typecheck/build can fail and the runtime default path can break.
- **Blocking:** latest head `20c6daebd` has failing CI. The combined CI job stops at `bun run fmt:check`, reporting formatting issues in `apps/server/src/git/Layers/GitManager.ts`, so lint/typecheck/tests/build did not get a chance to run.
- **Small / latent:** `apps/web/src/lib/gitReactQuery.ts` now sends `textGenerationModelSelection` to `summarizeDiff`, but the React Query key still ignores that selection. Since diff summaries are cached indefinitely, summaries can be reused across provider/model changes if this helper is used.

So: functionally this appears to be the right fix, but it needs the import fix, formatting fix, and ideally the cache-key guard before approval.

PR #314, opened 2026-07-08
Fix git status timeout outside repositories

Zker67:
## Summary
- Add a fast `git rev-parse --is-inside-work-tree` precheck before `statusDetails()` refreshes upstream state.
- Return the existing non-repository status immediately for non-repo cwd values, such as the Windows desktop default `C:\Users\Windows`.
- Treat only Git's expected exit-128 "not a git repository" result as a non-repository directory.
- Preserve missing-directory fallback behavior while propagating timeouts, dubious ownership, permission errors, and other unexpected Git failures as `GitCommandError`.
- Cover command ordering and error classification with regression tests.

## Why
On Windows desktop startup, Synara can ask for git status from a cwd that is not a repository. `statusDetails()` previously tried to resolve or refresh upstream state before determining that the cwd was not a repo, which could leave the UI stuck waiting on git status work that should have been skipped.

The precheck must remain narrow: unexpected Git failures inside real repositories must not be misreported as "not a repository."

## Validation
- Focused repository-precheck regressions: 4 passed.
- Full `apps/server/src/git/Layers/GitCore.test.ts`: 82 passed.
- `bun run --cwd apps/server typecheck`: passed.
- Affected-file `oxfmt`: passed.
- `git diff --check`: passed.
- Final diff against the PR base: 2 files (`GitCore.ts` and `GitCore.test.ts`).

### DIFF
--- a/apps/server/src/git/Layers/GitCore.test.ts
+++ b/apps/server/src/git/Layers/GitCore.test.ts
@@ -1507,6 +1507,30 @@ it.layer(TestLayer)("git integration", (it) => {
       }),
     );
 
+    it.effect("does not resolve upstream before rejecting non-repository directories", () =>
+      Effect.gen(function* () {
+        const operations: string[] = [];
+        const core = yield* makeIsolatedGitCore((input) =>
+          Effect.sync(() => {
+            operations.push(input.operation);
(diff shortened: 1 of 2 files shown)


Emanuele-web04:
@codex-bot Please review the latest head `a62277901c961e7559c263b336d705ec241473b2`.

PR #316, opened 2026-07-09
Fix runaway Claude usage: per-thread 1M window, compaction, fallback pinning, fewer restarts

Emanuele-web04:
## What Changed

This PR fixes several independent ways Claude sessions could consume far more subscription usage in Synara than in Claude Code CLI:

1. **The 1M context window is per-thread, never sticky.** Claude `contextWindow` is stripped from sticky composer state on write and from previously persisted sticky defaults on hydration. Choosing 1M once no longer silently enables it for every future thread.
2. **Compaction and usage warnings are explicit.** Claude SDK sessions always enable auto-compaction. A thread that starts on 1M now warns immediately that its compaction threshold is much larger, and prompts above 200k warn using plan-neutral usage language.
3. **Safeguard fallbacks stay on the effective model.** Real SDK `model_refusal_fallback` messages use snake_case fields; the adapter now decodes that wire shape, preserves the first refused model across repeated notifications, reports the actual serving model, and applies context-window changes to the fallback rather than flipping back.
4. **Fallback routing survives recovery.** The refused/effective model pair is stored in the Claude resume cursor, persisted immediately on `model.rerouted`, and restored across effort changes, idle runtime stops, app restarts, and other resume paths.
5. **Session replacement cannot orphan the old Claude runtime.** A replacement query is created first, then installed under a per-thread lifecycle lock after the previous query is closed. Failed replacement spawns leave the current session alive, and stale cleanup cannot delete a newer session.
6. **Claude restarts are limited to settings fixed at process spawn.** Model and context-window changes use in-session `setModel`; effort, thinking, fast mode, and ultracode still restart. The reactor tracks the selection actually applied to the live subprocess, including directly imported sessions and settings changed during an active turn.

Existing threads that intentionally or accidentally already use 1M keep their per-thread selection, but now receive an immediate warning telling the user how to switch that thread back to 200k. This avoids silently overriding intentional 1M choices while making inherited high-usage state visible.

## Why

Users reported simple Claude turns consuming 8–25% of their rolling usage window. Local session metadata showed Synara SDK prompts reaching roughly 300k–630k tokens without compaction, while Claude Code CLI sessions were much less likely to reach those sizes. A Fable refusal could also make the same large context pass through Fable and its Opus fallback in one turn, then repeat on later turns.

The dominant general cause was a one-off 1M choice leaking into future threads through sticky state. Fallback bouncing, unnecessary restarts, and unsafe same-thread replacement could further amplify usage or leave work running in an unreachable runtime.

## Verification

Focused verification on the latest head passes **359 tests**:

- Claude adapter, provider service, and command reactor: 175
- Shared model selection logic: 73
- Composer draft store: 89
- Traits picker browser suite: 22

Coverage includes the real snake_case fallback event, repeated fallbacks, fallback resume/restart, failed and successful same-thread replacement, actual-model turn reporting, pinned context changes, immediate 1M warnings, direct-start selection changes, mid-turn spawn-profile changes, sticky sanitization, and restart narrowing.

## Checklist

- [x] The runaway-usage causes are documented
- [x] Existing intentional per-thread 1M choices are preserved and warned
- [x] Production wire shapes and session lifecycle paths have regression coverage
- [x] Focused server, shared, web, and browser tests pass

---

> [!NOTE]
> **High Risk**
> Touches Claude session lifecycle, resume cursors, restart heuristics, and model fallback pinning—core paths that affect token usage, recovery after app restart, and which model actually serves a turn.
> 
> **Overview**
> Addresses runaway Claude subscription usage by changing how model selection, sessions, and composer defaults interact.
> 
> **Composer / sticky state:** Claude `contextWindow` is no longer written to sticky defaults (including hydration of old persisted state), so a one-off 1M choice stays on that thread only.
> 
> **Provider command reactor:** Tracks the selection actually applied to the live subprocess (`threadSessionModelSelections`), seeds from projection and `thread.created`, and uses shared `claudeSelectionRequiresRestart` so only spawn-fixed options (effort, thinking, fast mode, ultracode) restart Claude—not model or context-window changes (those use in-session `setModel`). Mid-turn metadata updates defer restart until the turn ends.
> 
> **Claude adapter:** Always enables SDK auto-compaction; warns on 1M thread start and when per-request prompt size crosses usage thresholds. Handles `model_refusal_fallback` SDK messages, emits `model.rerouted`, pins the session to the fallback model (with context applied on the fallback), persists the pair in the resume cursor, and restores it on resume. Session replacement is serialized with a per-thread lock, spawns the new query before retiring the old one, and cleans up failed installs without dropping a working session. Skips redundant `setModel` when unchanged.
> 
> **Provider service / ingestion:** Persists resume cursor immediately on `model.rerouted`; surfaces reroutes in the activity timeline.
> 
> **Shared:** New `claudeSelectionRequiresRestart` and moved `getEffectiveClaudeCodeEffort` for consistent restart rules.
> 
> **Version bump:** `0.4.1` → `0.4.2` in lockfile for affected packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 419daf984fb275439b3cc7282538587bc5698231. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -2337,6 +2337,12 @@ describe("ProviderCommandReactor", () => {
 
     await waitFor(() => harness.startSession.mock.calls.length === 1);
     await waitFor(() => harness.sendTurn.mock.calls.length === 1);
+    expect(harness.sendTurn.mock.calls[0]?.[0]).toMatchObject({
+      modelSelection: {
+        provider: "claudeAgent",
+        model: "claude-opus-4-7",
+      },
+    });
     harness.startSession.mockClear();
 
     // Context-window changes switch in-session via setModel on the next turn.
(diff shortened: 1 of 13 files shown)


Emanuele-web04:
@cursor review

Emanuele-web04:
@cursor review

Emanuele-web04:
@cursor review

Emanuele-web04:
@cursor review

PR #321, opened 2026-07-11
feat: cycle models with Alt+] / Alt+[ shortcuts

### LINKED ISSUE #283: [Feature]: Supports shortcut keys to quickly switch between commonly used models
### Before submitting

### Area

apps/desktop

### Problem or use case

This project is a Codex-like app where I frequently switch between models under the same provider (e.g. jumping between a fast/cheap model for simple edits and a larger model for harder reasoning, all within the same provider). Today switching requires opening the model picker with the mouse, scrolling the list, and clicking — repeated dozens of times a day. It's slow, breaks flow, and there's no way to jump straight to a "preferred" model.

### Proposed solution

Support keyboard shortcuts to cycle / switch between models within the currently selected provider:

A shortcut (e.g. Ctrl/Cmd+M or Alt+↑/↓) to cycle forward/backward through the models available under the active provider.
Optionally, Alt+<number> to jump to the Nth model in the provider's list, so a small set of frequently-used models can be reached in one keystroke.
The current model should be reflected in the status line / model indicator so the change is visible without opening the picker.

### Why this matters

Power users who work across multiple models per provider benefit most: reduced friction, hands stay on the keyboard, and mental context isn't broken by a UI round-trip. It also makes comparisons between models (e.g. A/B testing prompts) far more practical, encouraging better model selection per task.

### Smallest useful scope

A first pass only needs to support cycling (next/previous) through models of the current provider via a single shortcut, plus a visible indicator of the active model. No cross-provider switching, no global hotkey customization, no persisted "favorites" ordering — those can come later.

### Alternatives considered

_No response_

### Risks or tradeoffs

_No response_

### Examples or references

_No response_

### Contribution

- [ ] I would be open to helping implement this.

diliprt:
## Summary

- Adds `model.next` / `model.previous` keybindings (default **⌥ ]** / **⌥ [**).
- Cycles models for the **active provider only**, with **favorites first** then the rest of the discovered list.
- Updates the composer model selection without opening the picker.
- Surfaces the shortcuts in Settings → Keyboard Shortcuts.

Closes #283

## Motivation

Switching between commonly used models under the same provider currently requires opening the model picker every time. Maintainer guidance on #283 suggested next/previous commands with favorites-first order — this PR implements that v1.

## Implementation notes

| Piece | Detail |
|-------|--------|
| Contracts | `model.next`, `model.previous` commands |
| Defaults | `alt+]` / `alt+[` (web + server keybinding defaults) |
| Cycle order | favorites (stable storage order) → remaining options |
| Favorites storage | shared helper extracted from the model picker (`modelFavorites.ts`) |
| Handler | ChatView keyboard path (same gate and model-selection commit path as the picker) |

Follow-up review fixes:
- Enter the favorites-first list at the correct boundary when the previously selected model is no longer discovered.
- Normalize and deduplicate model/favorite slugs while ignoring stale favorites.
- Reuse the picker commit path so provider-specific cleanup, including collapsed Cursor variants, stays consistent.
- Stabilize browser coverage by retrying shortcut dispatch only until the hydrated key handler consumes the event; state assertions still fail normally if behavior is wrong.

Out of scope (intentionally): cross-provider cycling, `Alt+number` jump, custom keybinding UI.

## Before / after

**Before** — Keyboard Shortcuts only listed **Model picker** (⇧⌘M):

before

**After** — **Next model** and **Previous model** are listed:

after

## Validation

- [x] Cycle-order and unavailable-current edge-case unit tests
- [x] Web focused tests: 160 passed
- [x] Server keybinding tests: 29 passed
- [x] Contracts keybinding tests: 7 passed
- [x] Targeted Chromium integration: cycles next/previous without opening the picker
- [x] `bun fmt`
- [x] `bun lint` (0 errors; repository warnings remain)
- [x] `bun typecheck` (8/8 packages)
- [x] `git diff --check`
- [x] Terminal focus leaves the chords unhandled via `!terminalFocus`


Emanuele-web04:
@codex-bot review

Please review the latest head `864e3baa33140e957e5ec00569a51f2c9efdad6f`.

PR #356, opened 2026-07-14
fix(web): stop prototype members rendering as phantom file chips

### LINKED ISSUE #355: [Bug]: Inline code `constructor` / `__proto__` in a message permanently crashes the thread ("e.endsWith is not a function")
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Start the app (reproduces in the 0.5.3 desktop release and in bun run dev / bun run electron:dev on main).
2. Open any project and any thread.
3. Send a message containing this inline code span (backticks required):

```
what happens if a key is `constructor`
```

The app crashes as soon as the message renders.

All of these trigger it: 
```
`constructor`, `"constructor"`, `__proto__`, `"__proto__"`
```
 Surrounding quotes make no difference, because inlineCodeFilePath() strips them before the lookup.

It does not matter whether the text comes from the user or the assistant — both roles render through ChatMarkdown. An uploaded attachment named e.g. notes.constructor reaches the same defect through getAttachmentIconName().

### Expected behavior

`constructor` is ordinary prose in a JavaScript conversation. It should render as plain inline code (it is not a file path), and an unresolvable icon name should never take the app down.

### Actual behavior

The app unmounts and shows the error boundary: `Something went wrong. / e.endsWith is not a function.` The thread cannot be reopened afterwards — the crash reproduces on every render, so the conversation is effectively lost.

### Impact

Blocks work completely

### Version or commit

v0.5.3 (desktop release), also reproduces on main @ c254b102

### Environment

macOS 27.0  (arm64), Synara desktop 0.5.3, Bun 1.3.14, Node 24.17.0

### Logs or stack traces

```shell
MentionChipIcon.tsx:55 TypeError: name.endsWith is not a function
    at getCentralIconUrl (central-icons.tsx:34:31)
    at CentralIcon (central-icons.tsx:83:19)
The above error occurred in the <ForwardRef(CentralIcon)> component:
    <ForwardRef(CentralIcon)>   central-icons.tsx:83
    <MentionChipIcon>           MentionChipIcon.tsx:55
    <InlineMentionChip>         InlineMentionChip.tsx:43
    <OpenableFileChip>          ChatMarkdown.tsx:769
    <code>                      ChatMarkdown.tsx:1160
    <Markdown>                  ChatMarkdown.tsx:1255
    <ChatMarkdown>              MessagesTimeline.tsx:2561  (UserMessageBody)
React then unmounts the tree into CatchBoundaryImpl ("Warning: Error in route match: __root__/"), which is the full-app error screen.

For reference, the same crash from the packaged 0.5.3 release, where Ae is getCentralIconUrl and its caller is the CentralIcon forwardRef:

TypeError: e.endsWith is not a function
    at Ae (synara://app/assets/usePreviousValue-CVk5mVFU.js:1:37282)
    at synara://app/assets/usePreviousValue-CVk5mVFU.js:1:38239
    at To (synara://app/assets/main-DvMKKUH-.js:9:47526)
    at oc (synara://app/assets/main-DvMKKUH-.js:9:66676)
    at Nc (synara://app/assets/main-DvMKKUH-.js:9:82205)
```

[Screenshot 2026-07-14 at 18.18.11.png](https://github.com/user-attachments/assets/cf1ca3b5-a283-4be6-ab1c-8e4c37d01b9c)

### Workaround

_No response_

pray3m:
The icon tables were plain objects, so a lookup keyed on `constructor` or `__proto__` walked the prototype chain and returned an inherited member instead of an icon name. `getFileIconName` handed that function to `CentralIcon`, where `name.endsWith(".svg")` threw and unmounted the app through the error boundary.

The keys come from untrusted text, so a message containing the inline code `constructor` crashed the thread permanently: the throw reproduced on every render, leaving no way to reopen the conversation. Attachment names reached the same defect through `getAttachmentIconName`.

Back the tables with Maps, which never consult the prototype chain.

Fixes #355

## What Changed

The three file-icon tables in `apps/web/src/file-icons.ts` (`FILE_ICON_BY_BASENAME`, `FILE_ICON_BY_EXTENSION`, `FILE_ICON_BY_MIME_TYPE`) are now `Map`s instead of plain objects, and the seven lookups use `.get()` / `.has()`.

One production file, 33 lines. Regression tests cover the lookups (`file-icons.test.ts`) and the render that actually crashed (`ChatMarkdown.test.tsx`).

## Why

A message containing the inline code `` `constructor` `` or `` `__proto__` `` crashes the app with `TypeError: name.endsWith is not a function`, and the thread becomes **permanently unopenable** — the crash reproduces on every render, so there is no way back into the conversation.

The lookup keys are derived from untrusted text (chat content, attachment filenames), and a plain-object lookup walks the prototype chain:

1. `ChatMarkdown.tsx:742` — `inlineCodeFilePath()` strips wrapping quotes, so `"constructor"` becomes the bare key `constructor`.
2. `file-icons.ts:165` — `pathLooksLikeKnownFile()` evaluates `FILE_ICON_BY_BASENAME["constructor"]`, inherits `Object.prototype.constructor`, and treats the truthy result as a known file — rendering a file chip.
3. `file-icons.ts:200` — `getFileIconName()` returns that inherited **function** as the icon name.
4. `central-icons.tsx:34` — `getCentralIconUrl()` calls `name.endsWith(".svg")` on the function and throws mid-render, unmounting the tree into the error boundary.

`constructor` and `__proto__` are the live triggers because the basename is lowercased first (`toString` becomes `tostring` and misses by luck).

`Map` fixes this at the root: it never consults the prototype chain, so `getFileIconName`'s `=> string` signature is enforced by the data structure rather than assumed. Attachments reached the same defect through `getAttachmentIconName()` (an upload named `notes.constructor`), and that is fixed by the same change.

I deliberately left `inlineCodeFilePath()`'s quote-stripping alone — it is load-bearing (it is why `` `'src/foo.ts'` `` renders a chip) and it is not the bug. Narrowing it there would hide the markdown symptom while leaving the attachment path broken.

Reproduces on `main` and in the v0.5.3 release. I hit it in a normal chat about JS `Map`/`Set`, where the agent wrote a sentence about prototype pollution — and that message prototype-polluted the renderer.

Verified with `bun fmt`, `bun lint`, `bun typecheck`, and `bun run test`. The new tests fail without the `file-icons.ts` change and pass with it.

### Known sibling, not included

`AGENT_CHIP_COLOR_BY_NAME` (`apps/web/src/components/composerInlineChip.ts:123`) has the same unsafe-lookup shape, keyed on an agent's `color:` frontmatter. It is cosmetic rather than a crash — an agent colored `constructor` resolves to a function, `?? DEFAULT` does not catch it, and the chip renders unstyled. Left out to keep this PR focused; happy to send a follow-up.

## UI Changes

Before: sending a message containing `` `constructor` `` replaces the app with the error boundary, and the thread cannot be reopened.
Screenshot 2026-07-14 at 18 03 03

After: the token renders as ordinary inline code.
Screenshot 2026-07-14 at 19 11 09

## Checklist

- [ ] I included a video for animation/interaction changes

### DIFF
--- a/apps/web/src/components/ChatMarkdown.test.tsx
+++ b/apps/web/src/components/ChatMarkdown.test.tsx
@@ -293,6 +293,17 @@ describe("ChatMarkdown user variant", () => {
     expect(markup).not.toContain("Deep Research");
   });
 
+  it("keeps Object.prototype member names as literal inline code", async () => {
+    // `inlineCodeFilePath` strips wrapping quotes, so the quoted forms reach the icon
+    // tables as the bare keys `constructor` / `__proto__`.
+    for (const token of ["constructor", "__proto__", '"constructor"', '"__proto__"']) {
+      const markup = await renderUserMarkdown(`what if a key is \`${token}\``);
+
+      expect(markup).toContain("<code>");
+      expect(markup).not.toContain('data-slot="central-icon"');
+    }
+  });
+
   it("renders @-mention tokens as mention chips", async () => {
     const markup = await renderUserMarkdown("check @src/utils/model.ts please");
 
--- a/apps/web/src/file-icons.test.ts
+++ b/apps/web/src/file-icons.test.ts
@@ -1,6 +1,24 @@
 import { assert, describe, it } from "vitest";
 
-import { getAttachmentIconName, getFileIconName, inferEntryKindFromPath } from "./file-icons";
+import {
+  getAttachmentIconName,
+  getFileIconName,
+  inferEntryKindFromPath,
+  pathLooksLikeKnownFile,
+} from "./file-icons";
+
+// Object.prototype member names, which lookup keys can collide with because they are
+// derived from untrusted text. Only `constructor` and `__proto__` survive the lowercased
+// basename (`toString` becomes `tostring` and misses by luck), so pin them all.
+const PROTOTYPE_MEMBER_TOKENS = [
+  "constructor",
+  "__proto__",
+  "toString",
+  "valueOf",
+  "hasOwnProperty",
+  "map.constructor",
+  "obj.__proto__",
+];
 
 describe("getFileIconName", () => {
   it("uses exact filename matches from the Central mapping", () => {
@@ -56,6 +74,25 @@ describe("getFileIconName", () => {
     assert.equal(inferEntryKindFromPath("C:\\repo\\.gitignore"), "file");
     assert.equal(inferEntryKindFromPath("scripts"), "directory");
   });
+
+  it("never resolves an inherited Object.prototype member as an icon name", () => {
+    for (const token of PROTOTYPE_MEMBER_TOKENS) {
+      assert.equal(getFileIconName(token), "code-brackets", token);
+    }
+  });
+});
+
+describe("pathLooksLikeKnownFile", () => {
+  it("does not mistake an Object.prototype member for a known file", () => {
+    for (const token of PROTOTYPE_MEMBER_TOKENS) {
+      assert.isFalse(pathLooksLikeKnownFile(token), `${token} must not look like a known file`);
+    }
+  });
+
+  it("still recognizes real files", () => {
+    assert.isTrue(pathLooksLikeKnownFile("src/index.ts"));
+    assert.isTrue(pathLooksLikeKnownFile("package.json"));
+  });
 });
 
 describe("getAttachmentIconName", () => {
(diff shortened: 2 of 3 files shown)


Emanuele-web04:
@codex review

pray3m:
Heads up — v0.5.4 (`511a2eb5`) added a `typeof name !== "string"` guard in `getCentralIconUrl`, which fixes the **crash**. 🎉 I've rebased this on top of that.

I've also renamed this PR from "inline code constructor permanently crashes the thread" → "stop prototype members rendering as phantom file chips", since the crash is now handled upstream and that title no longer describes what this changes.

 This PR now targets the **root cause the guard sits downstream of**, which v0.5.4 doesn't address: `getFileIconName` still returns `Object.prototype.constructor` (a function) for the token `` `constructor` ``, so the word still renders as a clickable file-mention chip in ordinary JS chat — now iconless, and firing `console.error` on every render. Backing the lookup tables with `Map` means the token is never treated as a file: no phantom chip, no console spam, and `getFileIconName`'s `=> string` type stops being a lie.

 This also makes the v0.5.4 guard dead code — happy to remove it, or leave it as defense-in-depth, whichever you prefer. Regression tests included. 

cc @Emanuele-web04 

Before (v0.5.4) : 
Screenshot 2026-07-16 at 15 48 45


After (this PR) : 
Screenshot 2026-07-16 at 15 49 20

pray3m:
this issue is still relevant in v0.5.5

PR #358, opened 2026-07-14
Claude: fix rate-limit blowups; native CLI subagents & dynamic workflows

Bil0000:
## Summary

This PR fixes the critical usage-consumption bug in Claude sessions and brings the Claude provider to native Claude Code CLI parity for subagents and dynamic workflows: child threads, a composer picker strip, mid-run steering, foreground/background control, per-spawn model/effort selection, and a live workflow panel with phase navigation and per-agent metrics.

10 commits, structured to review in order. Every behavioral change is unit-tested against wire shapes captured from live runs of the real `claude` CLI (Agent SDK 0.3.154).

## Part 1 — Usage bug fixes (`5f139915`)

Users were losing up to ~10% of a 5h plan window on trivial messages. Three confirmed root causes, all fixed:

| Root cause | Fix |
| --- | --- |
| **Opus reroute pinning** — an Anthropic safeguard reroute on a cheap model pinned the session (and its resume cursor) to the Opus fallback for the rest of its life, silently billing every later turn at ~5× weight | Restore the user-selected model via the SDK `setModel` control request at turn completion; pinning removed from the resume cursor entirely. The `model.rerouted` warning event still fires per affected turn |
| **Uncapped transcript replay** — after any server/CLI restart, the next message re-sent the entire conversation as one uncached bootstrap message (effective cap 90k chars, head-biased truncation could drop the newest messages) | Bootstrap builder capped at 32k chars, newest-first retention, explicit omission notice; applies to stale-resume, handoff, fork, and sidechat paths |
| **Restart on thinking toggle** — flipping thinking restarted the subprocess with `resume`, re-ingesting full history | Applied live via the `applyFlagSettings` control request; no restart. (`effort`/`fastMode` keep restart semantics — the SDK exposes no control request for them) |

## Part 2 — Subagents (native CLI experience)

- **Child threads** (`dfef1fc3`): `forwardSubagentText` is enabled and messages tagged with `parent_tool_use_id` route through per-run scoped contexts into the existing subagent-thread machinery — each Task/Agent spawn becomes a real nested thread with its own transcript, turn lifecycle, and token meter (parent meter no longer absorbs subagent usage). Stop targets the individual task via `stopTask`, not the whole turn.
- **Picker strip** (`d8326efe`, `936334b8`, `997d0d2f`): strip above the composer lists live subagents (status dot, nickname, model · effort, status). Click to open a subagent thread; inside one, the strip persists — derived from the parent — with a "Main thread" row and the viewed agent highlighted, so you can hop between siblings or back. Includes a fix for the strip source being pre-filtered by the timeline's de-duplication (root-caused against real projection-DB payloads).
- **Steering a running subagent** (`b971d6ba`): typing in a subagent's composer queues a steer delivered through a PreToolUse hook as `additionalContext` on the agent's next tool call, with a delivery marker in the child transcript. (Raw injection via `parent_tool_use_id` was empirically proven to land on the main thread instead — the hook channel is the reliable path.) Steering a finished agent fails gracefully with a visible warning.
- **Foreground/background parity** (`b971d6ba`): per-row "Move to background" (the CLI's ctrl+b, via `backgroundTasks`) and "Stop" buttons; `is_backgrounded` state surfaces on rows; background agents that outlive the parent turn settle cleanly into their threads.
- **Per-spawn model + effort** (`b971d6ba`): the system-prompt append instructs the main agent to set the Agent tool `model` param and pick effort via new `worker-low/medium/high/xhigh` agent definitions (`AgentDefinition.effort`) — honoring explicit user requests verbatim, otherwise choosing per task complexity. Chosen model/effort render on strip rows.
- **Transcript trace** (`354d936e`): the parent transcript renders each subagent the way the CLI does — its last 4 tool calls inline, a "+N more tool uses" overflow line, live while running. Terminal rows now read Completed/Failed/Stopped instead of "Idle".

## Part 3 — Dynamic workflows (native CLI experience)

- **Workflow run card** (`cea827aa`, `2f436a62`): live card in the composer chrome with workflow name, running/total counts, ticking elapsed, stop button. Phases parsed statically from the script's `meta` literal (safe parser, no eval); live phase placement from `task_progress` `"<phase>: <label>"` descriptions.
- **Pause / resume / save** (`2f436a62`): pause = stop with a resumable card; resume re-invokes the Workflow tool with `{scriptPath, resumeFromRunId}` (completed agents replay from cache, so stop-then-resume ≈ pause); settled cards show the persisted `scriptPath · runId` with copy — the saved workflow. `runId`/`scriptPath`/`transcriptDir` come from the structured launch result, no text scraping.
- **Panel v2** (`64aa424d`): the phase rail is navigation — click a phase to see only its agents (auto-follows the current phase until a manual pick). Agent rows show label · model · effort · tokens · elapsed · status and expand inline to a detail view: status/model/effort line, tokens · tool calls · duration, clamped expandable prompt, recent tool names, and "Open thread" where applicable.
- **Live per-agent metrics**: the SDK stream carries no per-agent workflow data (verified against the CLI), so the adapter polls the run's `transcriptDir` every 2s while live — incremental byte-offset parsing of `journal.jsonl` and `agent-*.jsonl` yields per-agent model, tokens (formula verified against the CLI's own numbers), tool calls, and duration; the settled `output_file` provides the authoritative final backfill. Bounded: 5MB/file skip, 512KB/tick, change-only event emission, poller torn down on settle/stop.

## Follow-up round (from live manual testing) — `69c5a89b`, `ddc68c19`

- **Per-subagent Stop actually stops now.** Two-layer root cause found via the event store (zero interrupt events recorded = web-side break): the row dispatched an unresolved raw `toolu_…` id the server rejected silently, and the adapter's pre-`task_started` fallback *backgrounded* the task instead of stopping it. Fixed: stop dispatches the canonical `subagent:<parent>:<toolUseId>` thread id, and stops pressed before the task-id mapping arrives queue server-side and fire `stopTask` the moment `task_started` lands. `backgroundTasks` no longer appears anywhere on the stop path.
- **Stop-all button** in the strip header (shown when >1 agent running), fanning through the same single stop seam.
- **Worker-tier types hidden**: `worker-low/…` never render in nicknames, roles, or thread titles (they exist only to carry effort, which stays displayed as e.g. "Haiku · low"); already-persisted titles are cleaned at render.
- **Background action polish** (`4ab95bc4`): "Run in background" only offers on foreground running agents (already-backgrounded rows — whether patch-confirmed or spawned backgrounded — no longer show it; the gating previously keyed the patch set by a different id than the dispatch used), the icon is now a tray-drop glyph with a "Run in background (ctrl+b)" tooltip, and **Ctrl+B** backgrounds all foreground running subagents — native CLI parity.
- **Usage hardening round 2**: `effort` (non-max), `fastMode`, and `ultracode` selection changes now apply live via `applyFlagSettings` — only max-effort transitions still restart the session; stale resume retries the native cursor once before paying the capped bootstrap; `excludeDynamicSections` makes the system-prompt prefix statically cacheable; turns that ingest large uncached input (>50k tokens uncached, or <20% cache-read on >20k prompts) emit a visible once-per-session warning naming the likely cause; 30-minute idle reaper + benign-termination markers confirmed/enriched.

## New commands / contracts

- `thread.task.stop` and `thread.task.background` orchestration commands (decider → reactor → `ProviderService` → adapter), used by workflow stop/pause and subagent backgrounding.
- Provider seam `steerSubagent` mirroring the interrupt seam's subagent thread resolution.
- Schema-only contract extensions: task payloads carry workflow identity/phases/agent plans/runtime snapshots; `turn.steered` runtime event; subagent identity hints carry effort/background.

## Verification

- `bun fmt`, `bun lint` (0 errors; no new warnings introduced), `bun typecheck` (8/8) all pass.
- Full Vitest suite green across server / shared / contracts / desktop / scripts / effect-acp; web adds ~40 new test files/cases. The 5 failing web store-persistence suites (49 tests) fail identically on clean `main` — pre-existing, unrelated (verified in a throwaway worktree).
- Wire shapes validated by two live spike runs against the real CLI (subagent routing, hook correlation, workflow launch/progress/output-file shapes); UI flows manually tested end-to-end: subagent spawn → strip → thread switch → steer → background/stop, and workflow run → phase navigation → agent expand → pause/resume.

## SDK limitations & the workarounds shipped for each

Every limitation below is imposed by the Agent SDK wire protocol (empirically verified against the real CLI), and each ships with a working mitigation in this PR:

| SDK limitation | Workaround shipped |
| --- | --- |
| No input channel into a running subagent — messages sent with `parent_tool_use_id` land on the main thread (proven by live spike) | PreToolUse hook channel: steers queue server-side and deliver as `additionalContext` on the agent's next tool call, with a delivery marker in the child transcript. Residual gap: a purely-reasoning tail (no further tool calls) never receives it — steering a finished agent warns instead of hanging |
| No `pause` control request for tasks/workflows | Stop-then-resume: pause = `stopTask` + a resumable card; resume re-invokes `Workflow` with `{scriptPath, resumeFromRunId}`, and completed agents replay instantly from the run cache — functionally a pause. Paused-vs-stopped flag is session-local; after a reload the card reads "Stopped" but keeps its Resume button |
| Workflow phase structure never crosses the wire | Two-source reconstruction: the `meta` literal and per-agent `{label, phase, model, effort}` opts are statically parsed from the script (safe literal parser, no eval), and live phase placement is decoded from `task_progress` `"<phase>: <label>"` descriptions |
| No per-agent workflow metrics on the stream (workflow usage arrives aggregated; the per-agent `output_file` is written only at settle — confirmed by inspecting the CLI) | Bounded 2s polling of the run's `transcriptDir`: incremental parsing of `journal.jsonl` + `agent-*.jsonl` streams live per-agent model, tokens, tool calls, and duration; the settled output file backfills authoritatively. Live label↔agent matching is start-order heuristic until that correction; a server restart mid-run loses live metrics until settle |
| Agent tool input has a `model` param but no `effort` param | `worker-low/medium/high/xhigh` agent definitions (via `AgentDefinition.effort`) give the main agent a per-spawn effort lever that composes with the `model` param; system-prompt guidance wires up both automatic and user-directed selection |
| No control request for `effort`/`fastMode` session changes | Not worked around — these keep restart semantics (with the now-capped resume replay); model and thinking changes were moved to `setModel`/`applyFlagSettings` control requests so the common cases never restart |


Emanuele-web04:
@Bil0000 I reviewed this at head `4ab95bc4`. I don't think it's ready to merge yet; could you take a look at these findings?

### Merge blocker

1. **The authoritative final workflow snapshot can be skipped depending on SDK event order.** A terminal `task_updated` removes the task from `liveWorkflowTaskIds` and stops the poller ([ClaudeAdapter.ts](https://github.com/Emanuele-web04/synara/blob/4ab95bc48cbdb166b12f389112a2f1a59a23645e/apps/server/src/provider/Layers/ClaudeAdapter.ts#L3322-L3327)). The later `task_notification` only reads `output_file` while the ID is still in that set ([same file](https://github.com/Emanuele-web04/synara/blob/4ab95bc48cbdb166b12f389112a2f1a59a23645e/apps/server/src/provider/Layers/ClaudeAdapter.ts#L3593-L3601)). If `task_updated(completed|failed|killed)` arrives first, final per-agent models/states/metrics are silently lost. I suggest retaining workflow identity until `task_notification` and adding a regression test for that exact ordering.

2. **GitHub CI has not actually run.** The PR is currently `UNSTABLE`; only PR-size/vouch checks completed, while the CI run is `action_required` with no jobs. That should be approved/rerun before merge.

### Smaller correctness/reliability issues

- The contract documents `phaseIndex` as 1-based ([providerRuntime.ts](https://github.com/Emanuele-web04/synara/blob/4ab95bc48cbdb166b12f389112a2f1a59a23645e/packages/contracts/src/providerRuntime.ts#L494-L496)), but the UI fallback indexes `workflow.phases[phaseIndex]` directly ([WorkflowRunCard.logic.ts](https://github.com/Emanuele-web04/synara/blob/4ab95bc48cbdb166b12f389112a2f1a59a23645e/apps/web/src/components/chat/WorkflowRunCard.logic.ts#L602-L604)). When `phaseTitle` is absent, this maps to the next phase or “Other.”
- The settled workflow `output_file` is read without a size cap, even though the live transcript poller has a 5 MB guard. The final read should have equivalent protection.
- This is a very broad PR (45 files, +8.8k lines). Splitting usage-safety fixes from subagent/workflow feature work would reduce review and rollback risk, if practical.

### Verification I performed

- The PR merges cleanly with current `main` despite being four commits behind.
- 473 focused PR tests passed.
- The full merge-result package run was green apart from one local-image test caused by placing the review worktree under `/tmp`; the identical test passed after moving the worktree outside `/tmp`, confirming it was an environment artifact.
- I did not independently rerun fmt/lint/typecheck because the repository instructions prohibit those checks unless explicitly requested.

Once the workflow lifecycle issue is fixed and CI runs, this should be much safer to reassess.

Emanuele-web04:
@Bil0000 Follow-up: I fixed the actionable findings and pushed them directly to this PR.

- Preserved workflow identity through terminal `task_updated` so the later `task_notification` still backfills the authoritative final snapshot, with a regression test for that event order.
- Bounded settled workflow output reads to the same 5 MB safety limit as live workflow transcripts.
- Corrected the 1-based `phaseIndex` fallback and aligned its test fixtures.
- Fixed the PR's pre-existing brand-check test IDs.
- Merged current `main` and corrected the resulting keybinding-test formatting issue.

Fix commits include [c2d382a7a](https://github.com/Bil0000/synara/commit/c2d382a7a35b7a336013936daf5b7c7d50fe9fb7), [d4279edf5](https://github.com/Bil0000/synara/commit/d4279edf5823659dbc4501a3ca44e0ec508ec594), and [a437d5ea6](https://github.com/Bil0000/synara/commit/a437d5ea67ad3a37f705a1c90b630f74eb1f3b1b).

Final CI is green: release smoke, Windows regression, format, lint, typecheck, full tests, browser tests, desktop build, and preload verification all passed. [CI run](https://github.com/Emanuele-web04/synara/actions/runs/29528016926)

Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

PR #363, opened 2026-07-16
Fix stale Codex steering fallback

budah1987:
Fixes Codex steer submissions that race a completed turn.

Synara previously trusted projected session state when choosing `turn/steer`. If the terminal lifecycle update was delayed or missing, Codex rejected the request with `no active turn to steer` and the submitted message became a session error.

The reactor now verifies live provider state before steering, repairs stale projected state, and starts the submitted message as a normal turn when Codex is idle. It also rechecks after a steer failure and falls back if the active turn completed during the request.

Validation: `bun run --cwd apps/server test -- src/orchestration/Layers/ProviderCommandReactor.test.ts` (73 tests passed).


budah1987:
Closing: this change was published upstream by mistake and is intended to remain private.

PR #369, opened 2026-07-17
fix(shared): expose Claude effort ladder for custom model slugs

### LINKED ISSUE #367: [Feature]: Add effort level picker for custom Claude Code model slugs
### Before submitting

### Area

apps/desktop

### Problem or use case

When using custom model slugs for Claude code, it does not allow me to choose the effort level dynamically like the hard-coded built-in Claude models are.

For those using Claude Code with a custom proxy endpoint for other models, it would be nice to have this effort level picker since the native CLI allows for it.

It also seems that the  /workflow command doesn't work the same way as it does in the CLI. The workflows command is not appearing as an option in Synara. Is this something that could be looked at as well?

Additionally, the custom model slugs under Claude don't show up for the git writing feature.

### Proposed solution

Having the ability to choose effort level and even context window for custom model slugs and Claude code would be nice.

Also, having the ability to use workflows inside of Synara would be nice.

### Why this matters

Allows users to use Scenario to its full ability and Claude Code to its full ability while still using the models of their choice.

### Smallest useful scope

Being able to swap effort levels or use workflows gives users the ability to perform more heavy tasks using Claude code with other models.

### Alternatives considered

Manually switching effort level in the CLI

No workaround for workflows yet

### Risks or tradeoffs

No real risks or tradeoffs

### Examples or references

https://github.com/user-attachments/assets/a4369e13-cbd8-4f67-b85c-e71b7c04e38c

### Contribution

- [ ] I would be open to helping implement this.

diliprt:
## Summary

Custom / proxy **Claude Code** model slugs now get the same **effort (traits) ladder** as flagship built-in models, so the effort picker is no longer empty/hidden.

Closes #367

## Problem

`getModelCapabilities("claudeAgent", model)` only looked up the static catalog. Unknown slugs (custom proxies, alternate endpoints) resolved to **empty** capabilities → no effort levels in the traits picker, even though the Claude Code CLI accepts effort for those models.

## Solution

| Area | Change |
|------|--------|
| `getModelCapabilities` | For `claudeAgent` + unknown non-empty slug, fall back to flagship ladder (`claude-opus-4-8`) |
| Known models | Unchanged (Haiku still has no effort, Fable/Opus keep their ladders) |

Out of scope for this PR (can follow up): `/workflow` menu parity, custom slugs in git-writing model lists.

## Evidence

| Model slug | Before | After |
|------------|--------|--------|
| `claude-opus-4-8` | full ladder | unchanged |
| `claude-haiku-4-5` | no effort | unchanged |
| `my-proxy/custom-opus` | **empty** | flagship efforts (`low`…`ultracode`, default `high`) |

## Screenshots

before/after

## Test plan

- [x] `bunx vitest run packages/shared/src/model.test.ts` — **76 pass** (includes #367 custom-slug case)
- [ ] Manual: Settings → add custom Claude model slug → select it → traits picker shows effort levels

## Files

- `packages/shared/src/model.ts`
- `packages/shared/src/model.test.ts`


diliprt:
## Before / after screenshots

Custom Claude slug `my-proxy/custom-opus` — effort traits before (empty caps) vs after (flagship ladder):

#367 before/after

**Source:** deterministic UI proof of `getModelCapabilities("claudeAgent", customSlug)` behavior (PR #369).

diliprt:
Rebased onto latest `main` (clean). No functional changes.

PR #396, opened 2026-07-18
fix(server): recover Windows migrations and Codex startup

### LINKED ISSUE #393: [Bug]: Windows upgrade loops on SQLite backup fsync and blocks Codex recovery
## Before submitting

## Area

apps/server

## Steps to reproduce

1. On Windows, launch published Synara `v0.5.5` once so `state.sqlite` records migration 54 as `ProjectPullRequestPins`.
2. Upgrade that installation to current `main` (`ef16dee9` or any descendant of `af057220`), where migration 54 is `DurableProviderCommandDelivery` and the current high-water mark is 69.
3. Start Synara with the existing database.
4. Observe the pre-migration snapshot reach `FileHandle.sync()` after the temporary file was opened with `O_RDONLY` on Windows.
5. After correcting that handle locally, start or resume a Codex session while the primary Codex home contains SQLite databases and Synara's overlay contains recovered state or legacy SQLite-family links.
6. Observe overlay preparation and app-server initialization across a restart.

Minimal Windows filesystem repro:

```text
open("r")  + FileHandle.sync() -> EPERM
open("r+") + FileHandle.sync() -> success
```

## Expected behavior

- The v0.5.5 database is backed up once, migrations 54 through 69 reconcile successfully, and no incomplete snapshots remain.
- Codex SQLite databases, WALs, SHMs, and journals have one canonical home and are not assembled from independently managed per-file links.
- A recovered regular overlay database survives later starts; legacy SQLite links are removed without changing their source files.
- The app-server completes `initialize` followed by `initialized`, then can list models, read the account, start or resume the target thread, and accept a turn.
- A slow recovery-sensitive initialization has a bounded method-specific allowance without weakening normal request timeouts.

## Actual behavior

Three failures combine to make the Windows upgrade unusable:

1. **Migration durability blocker:** `MigrationBackup.ts` opens app-created backup, recovery-marker, and restore temporary files read-only before calling `sync()`. Windows returns `EPERM` before migration begins.
2. **Codex SQLite ownership violation:** overlay preparation deletes recovered regular SQLite files and relinks each database/sidecar from the primary Codex home when symlink creation succeeds. This can cross database generations and discard recovery progress.
3. **Initialization budget:** `initialize` shares the ordinary 20-second request timeout. An affected session reported `Timed out waiting for initialize.` while state recovery was in progress.

The desktop supervisor treats the backend exit as restartable. Ten fresh partial snapshots accumulated to 3,270,788,096 bytes in one reproduction because each deterministic retry created another full database copy and partial cleanup only considers artifacts older than 24 hours.

## Impact

Blocks work completely.

Existing Windows v0.5.5 installations are predictably exposed on the next upgrade because the recorded migration-54 identity diverges from current main. The failure prevents Synara from reaching a usable backend; repeated retries can also consume multiple gigabytes. When the backend does start, failure to complete the required Codex handshake or resume the thread breaks Synara's core conversation functionality.

OpenAI documents app-server as the interface for rich clients that need authentication, conversation history, approvals, and streamed events. The protocol requires `initialize` then `initialized` before other methods and uses `thread/resume` to continue stored conversations: https://learn.chatgpt.com/docs/app-server

OpenAI also documents that `CODEX_SQLITE_HOME` is the location for CLI and app-server SQLite state and defaults to `CODEX_HOME`: https://learn.chatgpt.com/docs/config-file/environment-variables#core-locations

## Version or commit

Published `v0.5.5` upgrading to `main@ef16dee9`; migration backup behavior introduced by `af057220`.

## Environment

Windows x64 desktop build; repository-pinned Bun 1.3.12; Codex CLI 0.144.5. The read-only `fsync` behavior was reproduced directly on Windows and in the packaged application.

## Logs or stack traces

```shell
Error: EPERM: operation not permitted, fsync
    at MigrationBackup-*.mjs

Error: Timed out waiting for initialize.
    at CodexAppServerManager.startSession

10 incomplete migration snapshots
3,270,788,096 bytes total
```

## Root causes and proposed correction

1. Use writable handles for the three app-created Windows durability sync points while preserving the Unix `O_RDONLY | O_NOFOLLOW` path.
2. Make SQLite-family state overlay-owned: preserve recovered regular files, remove legacy links, and skip source-home SQLite entries during overlay linking.
3. Give only `initialize` a 45-second timeout; retain 20 seconds for all other requests.

Required coverage includes normal upgrade and migration failure cleanup, recovered overlay state and legacy links, ordinary versus initialize timeouts, packaged startup from seeded v0.5.5 state, and a full `initialize` -> `thread/resume` -> harmless turn smoke.

## Workaround

There is no safe end-user workaround for the shipped upgrade. Stop Synara after the first deterministic failure to prevent restart-driven disk growth and preserve the finalized pre-migration backup. A local emergency build with the three corrections starts successfully, but users should not manually alter or restore SQLite files while any Synara or Codex process is running.

slashdevcorpse:
Fixes #393

## What Changed

This PR fixes one Windows recovery failure chain in three independently reviewable commits:

1. **Migration backup durability** (`29294143`)
   - Opens Synara-created temporary backup, marker, and restore files with read/write access before `fsync` on Windows.
   - Keeps the existing read-only, no-follow behavior for validation and Unix source reads.
   - Removes the current `.partial` artifact immediately if publication fails, so the next launch does not inherit a failed attempt.
   - Adds a v0.5.5-lineage migration regression through the current schema, including the historical ProjectPullRequestPins version-54 mismatch.

2. **Process-owned Codex SQLite recovery** (`5db8a602`)
   - Treats SQLite databases and their WAL, SHM, and rollback-journal sidecars as overlay-owned state instead of symlinking them back into the live source Codex home.
   - Preserves regular recovered overlay files across retries, removes legacy SQLite symlinks without following them, and leaves the source database untouched.
   - Pins both `CODEX_SQLITE_HOME` and the root Codex `-c sqlite_home=...` override to the per-process overlay. The CLI override is deliberately stronger than project, user, and system config; administrator-managed policy remains authoritative.
   - TOML-escapes the override and Unicode-escapes Windows batch metacharacters before invoking npm's `codex.cmd`.

3. **Recovery-specific initialize deadline** (`d3055257`)
   - Gives the mandatory app-server `initialize` handshake 45 seconds.
   - Keeps all ordinary JSON-RPC requests at the existing 20-second deadline.
   - Ensures success, timeout, and immediate transport-write failure all clear the pending request and timer.

## Why

On affected Windows upgrades, Synara could enter a deterministic restart loop:

1. migration-backup durability failed while syncing a Windows file descriptor;
2. Codex restart reused or linked the same mutable SQLite state instead of getting process-owned recovery state; and
3. Synara terminated the app server after the generic 20-second deadline while Codex was still recovering.

The result is not cosmetic degradation: the provider never completes initialization, existing threads cannot resume, and new Codex work cannot start. The three changes belong in one PR because they are consecutive stages of the same launch/recovery path, but each is isolated in its own commit and documented in its own PR comment for review and rollback.

## Happy and Failure Paths

| Part | Happy path retained | Failure/recovery path added |
| --- | --- | --- |
| Migration backup | Normal migration publishes a validated final backup and marker | Windows sync/publication failure removes the current partial; historical v0.5.5 schema reconciles without losing the pin |
| SQLite overlay | A regular recovered overlay database survives repeated preparation | Source DB/WAL/SHM/journal are never linked; legacy links are removed; inherited/configured SQLite locations cannot redirect the child |
| Initialize timeout | Fast initialization and normal requests behave as before | A recovering initialize may complete through 45s; it times out at 45s; non-initialize requests still time out at 20s; write failure rejects immediately |

## OpenAI Contract Checked

- [Codex app-server documentation](https://learn.chatgpt.com/docs/app-server): the client must send `initialize`, then `initialized`, before other methods; `thread/resume` continues an existing thread.
- [Codex environment-variable documentation](https://learn.chatgpt.com/docs/config-file/environment-variables#core-locations): `CODEX_SQLITE_HOME` is used by the CLI and app-server, while configured `sqlite_home` has precedence.
- [Codex 0.144.5 configuration source](https://github.com/openai/codex/blob/rust-v0.144.5/codex-rs/core/src/config/mod.rs): checked against the exact version used in the acceptance test.

## Verification

Passed on Windows PowerShell from the isolated worktree:

- Focused server suites: **104 passed, 2 skipped**.
- Shared Windows process suite: **22 passed**.
- `git diff --check origin/main...HEAD`.
- Real npm `@openai/codex@0.144.5` via `codex.cmd`: initialize, first harmless turn, stop/restart, exact `thread/resume`, and second harmless turn all passed.
- The live acceptance home deliberately included `R&D`, `100%`, spaces, and `^`; the resulting overlay SQLite DB/WAL/SHM files were regular files, not reparse points.

Broader local checks were attempted but do not provide a clean Windows baseline in the current checkout:

- Root `bun run test`: unrelated existing POSIX-path expectations in canary scripts and desktop profile/socket tests failed on Windows.
- Full server `bun run test`: unrelated existing CRLF, provider-health `.cmd` argument/casing, checkpoint newline, keybinding permission, and lifecycle-lock expectations failed; it was stopped before tests could exercise real provider state.
- `bun run release:smoke`: existing smoke harness stopped at `Expected a manual publication opt-in input.`
- `bun fmt`, `bun lint`, and `bun typecheck` were not run because repository instructions prohibit running them unless explicitly requested.

## Risk and Rollback

- Read/write access is limited to Synara-created temporary files that are immediately synced; source/validation reads remain read-only.
- SQLite ownership changes only the Windows-safe per-process Codex overlay; the user's source Codex state is not mutated.
- Managed Codex policy can still override the CLI flag by design; invalid duplicate root TOML assignments remain a separate configuration error.
- The longer deadline applies only to `initialize`, limiting the blast radius of a slow or wedged provider.
- Each stage can be reverted independently by reverting its corresponding commit.

## Checklist

- [x] Change is small and focused around one recovery chain
- [x] Root causes, behavior, and residual risks are explained
- [x] No UI changes; screenshots are not applicable
- [x] No interaction or motion changes; video is not applicable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,9 @@ jobs:
       - name: Test Effect Windows process spawn
         run: bun run --cwd apps/server test src/windowsProcessEffect.test.ts
 
+      - name: Test Windows migration backup and recovery
+        run: bun run --cwd apps/server test src/persistence/MigrationBackup.test.ts
+
   release_smoke:
     name: Release Smoke
     runs-on: ubuntu-24.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,9 @@ jobs:
       - name: Test Windows migration backup and recovery
         run: bun run --cwd apps/server test src/persistence/MigrationBackup.test.ts
 
+      - name: Test Codex SQLite overlay and initialization recovery
+        run: bun run --cwd apps/server test src/codexProcessEnv.test.ts src/codexAppServerManager.test.ts
+
   release_smoke:
     name: Release Smoke
     runs-on: ubuntu-24.04
--- a/apps/server/src/codexAppServerManager.test.ts
+++ b/apps/server/src/codexAppServerManager.test.ts
@@ -1070,6 +1070,99 @@ describe("startSession", () => {
   });
 });
 
+describe("sendRequest", () => {
+  function createRequestHarness() {
+    const manager = new CodexAppServerManager();
+    const context = {
+      nextRequestId: 1,
+      pending: new Map<string, unknown>(),
+    };
+    const writeMessage = vi
+      .spyOn(
+        manager as unknown as { writeMessage: (...args: unknown[]) => Promise<void> },
+        "writeMessage",
+      )
+      .mockResolvedValue(undefined);
+    const sendRequest = (
+      manager as unknown as {
+        sendRequest: (
+          context: unknown,
+          method: string,
+          params: unknown,
+          timeoutMs?: number,
+        ) => Promise<unknown>;
+      }
+    ).sendRequest.bind(manager);
+    const handleResponse = (
+      manager as unknown as {
+        handleResponse: (context: unknown, response: Record<string, unknown>) => void;
+      }
+    ).handleResponse.bind(manager);
+    return { context, handleResponse, sendRequest, writeMessage };
+  }
+
+  it("accepts an initialize response after the ordinary 20-second budget", async () => {
+    vi.useFakeTimers();
+    try {
+      const { context, handleResponse, sendRequest } = createRequestHarness();
+      const initializeRequest = sendRequest(context, "initialize", {});
+
+      await vi.advanceTimersByTimeAsync(30_000);
+      expect(context.pending.size).toBe(1);
+
+      handleResponse(context, {
+        jsonrpc: "2.0",
+        id: 1,
+        result: { ready: true },
+      });
+      await expect(initializeRequest).resolves.toEqual({ ready: true });
+      expect(context.pending.size).toBe(0);
+
+      await vi.advanceTimersByTimeAsync(15_000);
+      expect(context.pending.size).toBe(0);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("times out initialize at 45 seconds and ordinary requests at 20 seconds", async () => {
+    vi.useFakeTimers();
+    try {
+      const { context, sendRequest } = createRequestHarness();
+      const initializeRequest = sendRequest(context, "initialize", {});
+      const initializeRejection = expect(initializeRequest).rejects.toThrow(
+        "Timed out waiting for initialize.",
+      );
+
+      await vi.advanceTimersByTimeAsync(44_999);
+      expect(context.pending.size).toBe(1);
+      await vi.advanceTimersByTimeAsync(1);
+      await initializeRejection;
+      expect(context.pending.size).toBe(0);
+
+      const modelListRequest = sendRequest(context, "model/list", {});
+      const modelListRejection = expect(modelListRequest).rejects.toThrow(
+        "Timed out waiting for model/list.",
+      );
+      await vi.advanceTimersByTimeAsync(20_000);
+      await modelListRejection;
+      expect(context.pending.size).toBe(0);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("rejects immediately and clears pending state when the transport write fails", async () => {
+    const { context, sendRequest, writeMessage } = createRequestHarness();
+    writeMessage.mockRejectedValueOnce(new Error("injected transport write failure"));
+
+    await expect(sendRequest(context, "initialize", {})).rejects.toThrow(
+      "injected transport write failure",
+    );
+    expect(context.pending.size).toBe(0);
+  });
+});
+
 describe("sendTurn", () => {
   it("sends text and image user input items to turn/start", async () => {
     const { manager, context, requireSession, sendRequest, updateSession } =
--- a/apps/server/src/codexAppServerManager.test.ts
+++ b/apps/server/src/codexAppServerManager.test.ts
@@ -8,12 +8,14 @@ import {
   readFileSync,
   readlinkSync,
   rmSync,
+  symlinkSync,
   writeFileSync,
 } from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { PassThrough } from "node:stream";
 import { ApprovalRequestId, ThreadId } from "@synara/contracts";
+import { prepareWindowsSafeProcess } from "@synara/shared/windowsProcess";
 
 import {
   buildCodexProcessEnv,
@@ -21,6 +23,7 @@ import {
   resolveCodexBrowserUsePipePath,
 } from "./codexProcessEnv";
 import {
+  buildCodexAppServerArgs,
   buildCodexInitializeParams,
   CODEX_DEFAULT_MODE_DEVELOPER_INSTRUCTIONS,
   CODEX_PLAN_MODE_DEVELOPER_INSTRUCTIONS,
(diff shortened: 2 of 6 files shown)


slashdevcorpse:
### Part 1/3 — Windows migration backup durability

**Commit:** [`29294143`](https://github.com/Emanuele-web04/synara/pull/396/commits/29294143b63a8a6dbff263bd24baf4ddcb97c662)

**Root cause and usability impact:** the migration publication path synced Synara-created temporary files through a descriptor mode that is not reliably flushable on Windows. When that failed, the current `.partial` could remain, every desktop restart retried the same upgrade boundary, and Synara could not reach provider startup.

**Adjustment:** backup, marker, and restore temporary files are opened read/write only for their create/write/sync lifecycle. Existing validation and Unix source reads remain read-only/no-follow, and the exact current partial is removed immediately when publication fails.

**Happy path:** a v0.5.5-lineage database (including the historical ProjectPullRequestPins schema recorded at migration 54) migrates through the current version, publishes one validated backup and marker, and preserves the sentinel pin.

**Non-happy path:** an injected open/sync/publication failure rejects the migration and immediately removes that attempt's partial artifact, leaving the next launch a clean retry instead of another stale-artifact loop.

Focused migration/restore coverage passes on Windows and the new workflow step runs this boundary on `windows-latest`.

slashdevcorpse:
### Part 2/3 — Process-owned Codex SQLite recovery

**Commit:** [`5db8a602`](https://github.com/Emanuele-web04/synara/pull/396/commits/5db8a6020343544c6fca05298efbc01a5c5c251c)

**Root cause and usability impact:** the Windows-safe overlay could symlink Codex SQLite databases and WAL/SHM/journal sidecars back to mutable source state. Inherited `CODEX_SQLITE_HOME` or a configured `sqlite_home` could also redirect the app-server away from the overlay, so restarting Synara repeatedly reopened the same recovery state and blocked both thread resume and new work.

**Adjustment:** SQLite families are overlay-owned regular files. Preparation preserves regular recovered overlay state, removes legacy SQLite symlinks without following them, skips source SQLite links, and pins both the environment and root CLI `-c sqlite_home=...` value. The CLI TOML value is escaped safely through npm's Windows `codex.cmd`; administrator-managed Codex policy intentionally remains authoritative.

**Happy path:** repeated preparation keeps an already-recovered regular overlay database and resumes the exact provider thread without touching source SQLite state.

**Non-happy path:** legacy DB/WAL/SHM/journal links are removed, inherited/project/user/system SQLite locations cannot redirect the child, and Windows batch metacharacters are encoded before crossing the `.cmd` boundary.

The real npm `@openai/codex@0.144.5` passed initialize, first turn, restart, exact `thread/resume`, and a second turn from a home containing `R&D`, `100%`, spaces, and `^`. This follows OpenAI's [SQLite-home precedence](https://learn.chatgpt.com/docs/config-file/environment-variables#core-locations) and [app-server resume contract](https://learn.chatgpt.com/docs/app-server).

slashdevcorpse:
### Part 3/3 — Recovery-specific initialize timeout

**Commit:** [`d3055257`](https://github.com/Emanuele-web04/synara/pull/396/commits/d3055257783da39e0da850b4f7d6c14f476b0de8)

**Root cause and usability impact:** `initialize` shared the generic 20-second JSON-RPC deadline. A healthy Codex process doing Windows SQLite recovery could therefore be killed before completing the mandatory handshake, producing the reported “Timed out waiting for initialize” loop and making the provider unusable.

**Adjustment:** only `initialize` receives a 45-second deadline; ordinary requests remain at 20 seconds. Pending entries and timers are cleared on success, timeout, and immediate transport-write failure.

**Happy path:** normal initialization completes immediately with the same protocol sequence, and ordinary requests retain their existing timeout.

**Non-happy path:** delayed initialization remains pending after 30 seconds and may succeed before 45; it rejects at 45 seconds if still wedged. A non-initialize request still rejects at 20 seconds, while a write failure rejects immediately rather than waiting for either deadline.

Fake-timer and transport-error coverage passes, and the live Codex 0.144.5 acceptance completed both initialization handshakes. The boundary matches OpenAI's documented requirement to send [`initialize`, then `initialized`, before other app-server methods](https://learn.chatgpt.com/docs/app-server).

PR #431, opened 2026-07-19
feat(web): add optional advanced effort slider

xFurti:
## What changed

Adds an optional direct-manipulation effort slider to the composer.

- Adds the **Advanced effort slider** setting under Runtime behavior.
- Keeps the existing effort radio list as the default experience.
- Shows a compact slider card with provider-defined effort levels, a current-level indicator, keyboard support, and stable layout.
- Uses local preview state while dragging and persists the selected effort when the interaction completes.
- Falls back to the standard picker for provider variants and Claude's prompt-controlled Ultrathink mode.
- Preserves the existing fast-mode control.

## Why

Models with multiple reasoning levels benefit from a faster, more visual way to tune effort, without changing the default experience for users who prefer the existing list.

## Screenshots

Screenshots of the composer control and its opt-in setting are attached below.

## Validation

- `bun run test -- src/appSettings.test.ts` — 56 passed
- `bun run test:browser -- src/components/chat/TraitsPicker.browser.tsx` — 24 passed

### DIFF
--- a/apps/web/src/appSettings.test.ts
+++ b/apps/web/src/appSettings.test.ts
@@ -806,6 +806,7 @@ describe("AppSettingsSchema", () => {
       enableAppSnap: false,
       appSnapPlaySound: true,
       enableAssistantStreaming: true,
+      enableAdvancedEffortSlider: false,
       sidebarProjectSortOrder: DEFAULT_SIDEBAR_PROJECT_SORT_ORDER,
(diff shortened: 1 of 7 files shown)


xFurti:
Here’s a preview of the optional effort slider. It can be enabled in Settings → Behavior, so users who prefer the current selector can keep it unchanged.
settings
effort

PR #434, opened 2026-07-20
fix: honor autoCompactWindow when projecting configured Claude context window

### LINKED ISSUE #433: `1M` context window selection is silently ignored — `session.configured` writes `autoCompactWindow` but the ingestion reads `contextWindow`
## Summary

Selecting **1M** for a Claude thread has no effect: the context meter stays at the 200k default (shown as ~167k after the reserved budget), even on a fresh session with 1M explicitly selected. The cause is a field-name mismatch between the code that emits `session.configured` and the code that consumes it.

## Root cause

`context-window.configured` — the activity that locks the meter to the selected window — is produced in exactly one place, `buildConfiguredContextWindowPayload` (`apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts`), which reads **`config.contextWindow`**:

```ts
const configuredContextWindow = asString(config?.contextWindow)?.trim().toLowerCase();
const maxTokens =
  asPositiveFiniteNumber(config?.contextWindow) ??
  (configuredContextWindow === "1m" ? 1_000_000 : configuredContextWindow === "200k" ? 200_000 : undefined);
if (maxTokens === undefined) return undefined; // → activity never emitted → meter never locks
```

But `ClaudeAdapter` (`apps/server/src/provider/Layers/ClaudeAdapter.ts`) emits that event with the window under **`config.autoCompactWindow`**:

```ts
type: "session.configured",
payload: { config: { ...(requestedAutoCompactWindow ? { autoCompactWindow: requestedAutoCompactWindow } : {}) } },
```

`config.contextWindow` is therefore always `undefined`, the activity is never produced, and the meter falls back to whatever window the SDK reports.

## How it regressed

`c865c5e8` renamed the window from `contextWindow` to `autoCompactWindow` in the adapter but left `buildConfiguredContextWindowPayload` reading `contextWindow`:

| | producer (`ClaudeAdapter`) | consumer (`buildConfiguredContextWindowPayload`) |
|---|---|---|
| before `c865c5e8` | `{ contextWindow }` | reads `contextWindow` ✅ |
| `c865c5e8` → `HEAD` | `{ autoCompactWindow }` | reads `contextWindow` ❌ |

The existing test still feeds `contextWindow: "1m"` (matching the consumer, not the real producer), so the regression stayed green in CI.

## Visibility

When the SDK reports the model's full 1M window, the wrong fallback happens to be correct and the bug is hidden; when it reports the historical 200k window for a native-1M model, the 1M selection is silently lost.

## Fix

Read `config.autoCompactWindow` first, falling back to `config.contextWindow`. PR incoming.

keepure:
## What

`buildConfiguredContextWindowPayload` now reads the configured window from `config.autoCompactWindow` first, falling back to `config.contextWindow`. This restores the `context-window.configured` activity for Claude threads, so selecting **1M** actually locks the context meter to 1M again.

Fixes #433

## Why

`ClaudeAdapter` emits `session.configured` with the selected window under `autoCompactWindow` (renamed from `contextWindow` in `c865c5e8`), but the ingestion still read `contextWindow`. The lookup was therefore always `undefined`, the `context-window.configured` activity was never produced, and the meter fell back to whatever window the SDK reported — capping 1M threads at the default. Full analysis in #433.

## Changes

- **`ProviderRuntimeIngestion.ts`** — read `config.autoCompactWindow ?? config.contextWindow`. The emitted payload keeps the `contextWindow` label the web meter already consumes, so the frontend contract is unchanged.
- **`ProviderRuntimeIngestion.test.ts`** — added a regression test that emits `session.configured` with `autoCompactWindow: "1m"` and asserts the `context-window.configured` activity is produced with `maxTokens: 1_000_000`. This test fails before the fix.

## Compatibility

Backward compatible: events that still carry `contextWindow` keep working, and the payload continues to expose the `contextWindow` label unchanged.


keepure:
Hi @Emanuele-web04  — small one-line contract fix for #433: the Claude adapter emits the selected window as autoCompactWindow, but the ingestion still reads contextWindow, so a 1M selection silently falls back to the default. The change reads autoCompactWindow ?? contextWindow (backward compatible) and adds a regression test. Would appreciate a look whenever you have a moment — happy to adjust anything. Thanks!

Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

Emanuele-web04:
@codex review

PR #438, opened 2026-07-21
fix(server): respect --home-dir in mcp subcommands

horizzon3507:
## Problem
`mcp serve` and `mcp pair` ignored `--home-dir` and always used `~/.synara`.

## Root cause
Both the root `synara` command and the `mcp` subcommands defined a `--home-dir` flag. The Effect CLI assigns duplicate flags to the parent command, so the subcommand never received the value.

## Fix
- Removed the duplicate `--home-dir` flag from `mcp serve` and `mcp pair`
- `mcp` subcommands now read `synaraHome` from the parent command via `yield* baseServerCommand`
- Restructured `serverCommand` into `baseServerCommand` so subcommands can reference it before it is fully assembled

## Verification
```bash
synara mcp serve --home-dir /home/gabriel/.synara-canary
```
Now resolves to the provided home directory instead of defaulting to `~/.synara`.

### DIFF
--- a/apps/server/src/main.test.ts
+++ b/apps/server/src/main.test.ts
@@ -193,6 +193,20 @@ it.layer(testLayer)("server CLI command", (it) => {
     }),
   );
 
+  it.effect("passes the root --home-dir flag to MCP subcommands", () =>
+    Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


Emanuele-web04:
@codex review

PR #125, opened 2026-05-19
feat(providers): add Hermes ACP provider

kartikkabadi:
## Summary

- Adds **Hermes Agent** (`hermes acp`) as a ninth provider using the existing ACP subprocess stack (spawn, optional auth method selection, permissions, turns, resume).
- Extends `AcpSessionRuntime` for conditional authenticate and in-session `setSessionModel`.
- Wires contracts, server registry/health, and web settings/composer (static `hermes-agent` model + `customHermesModels`).

**Deferred (documented in `HermesAdapter.ts`):**
- **Runtime model list** — Hermes ACP docs cover `session/set_model` but no public `list_models` RPC; MVP stays static + custom slugs (`supportsRuntimeModelList: false`).
- **User-input elicitation** — Hermes editor UX uses ACP permission/terminal auth, not AskUserQuestion-style prompts; `respondToUserInput` remains unimplemented.
- **Health auth status** — probes use `hermes acp --version` / `--check` only; `authStatus` stays `unknown` (no machine-readable `hermes status` output).

Includes a small `PiAdapter.test.ts` fix for `exactOptionalPropertyTypes` (pre-existing on `main` after PR #124).

## Test plan

- [x] `bun run test` (targeted):
  - `apps/server/src/provider/hermesAcp.test.ts`
  - `apps/server/src/provider/Layers/HermesAdapter.test.ts`
  - `apps/server/src/provider/Layers/ProviderHealth.test.ts`
  - `packages/contracts/src/orchestration.test.ts`
  - `apps/web/src/appSettings.test.ts`
- [x] `bun fmt`, `bun lint`, `bun typecheck`
- [x] Isolated dev dry-run: `bun run dev -- --home-dir ./.dpcode-hermes --port 58090 --dry-run`
- [x] Host CLI: `hermes acp --check` OK (v0.14.0 on PATH)
- [ ] Manual UI (requires `hermes acp` credentials): health badge, first turn, permissions, custom model slug, session resume
- [ ] Attach `.scratch/hermes-qa-*.png` after manual pass on synced build

## Hermes install

```bash
pip install -e '.[acp]'   # or ensure `hermes` on PATH
hermes acp --version
hermes acp --check
hermes model              # configure provider credentials
```

Docs: https://hermes-agent.nousresearch.com/docs/user-guide/features/acp

Made with [Cursor](https://cursor.com)

PR #110, opened 2026-05-13
[codex] Add desktop Remote SSH support

frizynn:
## Summary

Adds desktop Remote SSH support so the local Electron app can connect to a DP Code server started on an SSH host, tunnel the backend WebSocket/HTTP connection, and open remote projects from the local UI.

## Details

- Adds desktop IPC/contracts/preload APIs for Remote SSH connection status and local clipboard image reads.
- Starts the remote DP Code backend over SSH with key-based, non-interactive options, remote project cwd, health readiness, stderr diagnostics, and lifecycle cleanup.
- Adds localhost forwarding for the built-in browser, including HTTP/HTTPS/WS/WSS loopback URLs, fail-closed behavior on forwarding errors, and logical URL preservation in the address/history state.
- Adds local clipboard image paste fallback for Electron so pasted screenshots/images still attach while connected to a remote backend.
- Adds focused tests for remote startup command construction and readiness handling.

## Validation

- `git diff --cached --check`
- `npx -y -p node@24.13.1 -p bun@1.3.9 -c 'cd apps/desktop && bun run test'`
- `npx -y -p node@24.13.1 -p bun@1.3.9 -c 'cd apps/web && bun run test'`
- `npx -y -p node@24.13.1 -p bun@1.3.9 -c 'bun run build:desktop'`
- Local SSH smoke for the remote backend manager and browser port forwarder during implementation

PR #109, opened 2026-05-13
Add Pi provider integration

isthatyousaf:
## What Changed

Adds first-class Pi harness support as a new DP Code provider.

- Adds `pi` to provider/model/settings/runtime contracts and compatibility handling.
- Adds a direct Pi SDK adapter using `@earendil-works/pi-*` packages.
- Wires Pi into provider discovery, health, runtime registry, session directory, model selection, settings, and thread handoff flows.
- Adds Pi UI branding via the supplied Pi icon in the sidebar, chat header, provider picker, plugin library, and settings.
- Supports Pi dynamic/custom models without hardcoded fallback behavior, including provider-qualified custom slugs such as `openai/gpt-5.5`.
- Supports Pi skill discovery/autocomplete using native `/skill:<name>` commands.
- Adds Pi `/compact` feedback and Pi-only `/reload` command handling without sending those commands to the model.
- Surfaces Pi runtime failures as visible provider runtime errors and failed turns instead of silent in-turn failures.
- Adds regression coverage for Pi model selection, settings, projection/replay behavior, contract decoding, and UI state handling.

## Why

DP Code already has a provider abstraction, and Pi is a good fit as an unopinionated coding-agent harness. This PR integrates Pi directly through its SDK rather than wrapping it through a subprocess protocol.

The adapter keeps Pi-specific behavior in the provider layer where possible, so shared orchestration/UI code continues to consume normal DP Code runtime events and model-selection shapes.

## UI Changes

Pi now appears anywhere provider identity matters:

- Provider/model picker
- Chat header and sidebar thread branding
- Settings provider controls
- Plugin/provider surfaces
- Skill autocomplete loading/status states
- Context-window meter labels for Pi token usage

No screenshots were captured in this headless validation environment.

## Test Results

- `bun fmt && bun lint && bun typecheck && bun run test` — passed
- `bun typecheck` — passed after the final settings-schema polish
- `bun run dist:desktop:linux` — passed and produced `release/DP-Code-0.0.44-x86_64.AppImage`

## Review Focus

Please focus review on:

- `apps/server/src/provider/Layers/PiAdapter.ts` for Pi SDK lifecycle/event mapping.
- Model-selection preservation for Pi custom/dynamic models.
- Projection/replay handling for provider-qualified turn-start model updates.
- Runtime error surfacing from Pi prompt/reload failures.
- UI provider branding and picker behavior when Pi is selected.

## Checklist

- [ ] This PR is small and focused
- [x] I explained what changed and why
[...]

PR #589, opened 2026-08-08
Add provider limits and actual usage history

kartikkabadi:
## What this does

- Keeps account limits and observed activity as two separate usage planes, with all Synara providers visible and explicit signed-out / unsupported states.
- Adds a machine-activity contract: 24h / 7d / 30d token and session totals, token dimensions, measured vs recorded cost, model and upstream-provider breakdowns.
- Reads Codex and Claude local archives plus OpenCode and Grok provider-owned SQLite / session history. Reads are bounded, read-only, parameterized, deduplicated; no credentials read, no telemetry sent.
- Shows Synara-observed activity alongside provider account limits consistently in Usage & limits, the provider menu, and profile stats (with token provenance labels).
- Documents the CodexBar / OpenUsage research and attribution in `docs/provider-usage-research.md`; no provider code copied.
- Rebased onto upstream/main anchor `562c5fea`. Against that anchor the Kilo provider is retired (legacy rows migrate to OpenCode via `LEGACY_PROVIDER_MIGRATIONS`); this lane drops the PR's Kilo fetcher/tests/docs claims, retargets coverage to OpenCode, and adds Grok machine activity plus a per-provider usage snapshot query.

## Behavior change

- Usage & limits screen: per-provider snapshot query (`serverProviderUsageSnapshotQueryOptions`, 30s stale/refetch, no retry) feeds the settings panel, activity card, and menu model.
- `unsupported` live snapshots are honest capability results, not failed fetches: they no longer hide runtime limits, local activity, or thread-reported limits (`resolveProviderUsageSummary` only treats `needs-auth` / `error` as blocking).
- Profile model-usage aggregates per-message `modelUsage` turns (deduped by turn id) alongside cumulative totals; OpenCode/Kilo cumulative rows stay attributed through the shared migration map.
- Cursor `turn.completed` payloads carry the persisted ACP session id so cost baselines reset across same-provider sessions.
- Contracts are purely additive and optional (`activity` on the usage snapshot, `turnCount`/`threadCount`/`costUsd`/`upstreamProviderId` on profile token usage); older consumers still decode.

## Reviews

- Ponytail-ultra pass over `gh pr diff`: no dead code kept (Kilo fetcher, tests, and copy removed or retargeted); no new abstractions; smallest branch-based optionals where the type system allows.
- Thermos lens 1 (bugs/security): SQLite opens read-only with parameterized queries and never throws into the refresh path; `codexHomePath` is server-validated settings, never a client path; `auth.json` discovery prefers the configured Codex home so live and local read the same account. No blocking findings.
- Thermos lens 2 (maintainability): fixed the `unsupported`-blocks-fallback regression, an oxfmt long line, and `no-map-spread` warnings in the new mappers (branch returns instead of conditional spreads). Conditional spreads remain only where `exactOptionalPropertyTypes` + readonly Schema types require omission rather than explicit `undefined`. Fixed and pushed; nits below.

## Tests

- `bun run fmt:check`: clean on all 2937 files.
- Server typecheck (`tsc --noEmit`) and web typecheck: clean.
- Scoped server suites (`providerUsage`, `providerUsageSnapshot`, `profileStats`): 19 files, 166 tests pass.
- Scoped web suites (`serverReactQuery`, `openUsageRateLimits`, `ProviderUsageSettingsPanel`, `useProviderUsageSummary`): 4 files, 31 tests pass.
- Headed live proof (slot-approved): dev server :4362 + web :6322, Usage surfaces exercised, >=2 screenshots + Playwright trace captured to local `evidence/589/` (kept untracked, never posted); servers stopped and ports verified DOWN, slot released. OS screen recording is TCC-blocked, so no .mov is claimed.
- Full CI (Format, Lint, Typecheck, Test, Browser Test, Build + Windows Regression + Release Smoke) re-run on the pushed head; status in Merge note.

## Socket surface

None. `gh pr diff` grepped for `ws|socket|gateway|transport` hits only model-row and Kilo-migration comments. No live connect/send/receive or killed-server path to exercise.

## Follow-ups

- `cursorTurnProviderThreadId` (CursorAdapter) returns `{...} | {}` spread into payloads; could become an explicit optional field when that payload type is next touched.
- The pre-existing `Sidebar.import` smoke-test 15s timeout under load is unrelated to the usage files and still stands as a known test-environment limitation.
- Kilo historical references remain only in migration maps, changelog/whats-new history, and provider docs noting the retirement.

## Merge note

- DRAFT; merge-ready lane for PR 589 onto anchor `562c5fea`. Do not merge from this lane; upstream merge decision stays with the PR owner.
- CI: `MERGEABLE`, awaiting green on head `6fee0d652` (rebased, force-pushed with lease to the PR's own fork branch `codex/track-usage-limits-and-tokens`; no other branches touched, no comments posted).

## Gauntlet handoff (2026-09-04)

**Status: NOT ready — 0 BLOCKER, 8 MAJOR, 6 MINOR.** Review: 6 lenses + oxlint anti-slop (54 hits on changed lines). Full reports held locally by Kartik.

**Must-fix before merge**
1. **Name-sniffed math.** Token/counter math branches on provider-name strings (`profileStatsArchive.ts:318-330`, live mirror `profileStats.ts:1088-1160`) — a cumulative-counter adapter on a new provider inflates lifetime totals hugely. Fix: mark counter semantics per row at write time, delta-math fallback when unknown.
2. **Unbounded growth.** Deleted-thread tables grow forever; every history read full-scans + JS-filters 90d. Fix: SQL date floor (90-day + skew) + daily rollups.
3. **FAIL fix:** future-dated archive events inflate machine 24h/7d/30d + `updatedAt` (profile path already excludes them — mirror the guard).

**Also queued:** 7 CUT (~88 lines: never-written `estimatedCostUsd`, two test-forwarder shims, noise hunks) + 5 SLIMs. Big new files are LOAD-BEARING — fat is seams + dup tests + one dup boolean only.

**Remaining:** Fix → reverify → websocket + playwright E2E → rebase onto `562c5fea` → CI green → undraft.


kartikkabadi (commented):
Two correctness issues remain in the new provider-usage accounting.

**P1 — mixed telemetry can silently drop historical tokens.** `buildProviderUsage()` computes `providersWithModelUsage` at provider scope. If even one `turn.completed` row for a provider contains `modelUsage`, every `queryTokenActivity()` row for that provider is skipped. That means an existing installation with older turns that only have `context-window.updated` token counters will lose all of those historical tokens as soon as one newer turn starts reporting `modelUsage`. The suppression needs to be scoped to the specific turn/day/model evidence being replaced, or the two sources need a deduplication key rather than a provider-wide switch.

**P1 — the first cumulative cost in every thread is discarded.** In `queryProviderTurnActivity()`, a `cumulativeCostUsd` row is only converted to a cost when `previous_cumulative_cost_usd IS NOT NULL`. The first cumulative sample therefore contributes `NULL`, so a thread with cumulative values `1.25 -> 3.75` is reported as `$2.50` instead of `$3.75`. The new regression test currently codifies that undercount. When there is no previous cumulative sample, the first non-negative cumulative value should contribute its full value (while still preserving the reset handling for later decreases).

Exact-head CI for `f513f9e9f53167a6dc3511e34743de307423278f` is still in progress.

Review #102
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Review of the new exact head confirms the earlier provider-wide token suppression problem has been narrowed, and the configured Codex-home path is now server-owned and consistently applied. Two accounting correctness issues still remain, both around cost completeness.

1. The first cumulative cost sample in a session is still discarded. A one-turn Cursor/ACP session can therefore report no cost at all, and a multi-turn session undercounts by its first sample.
2. Cost totals still have no coverage signal. If some completed turns report cost and others do not, the UI can present the partial sum as an ordinary complete total.

These are material because the feature explicitly promises honest actual usage rather than fabricated or silently incomplete totals. I would not treat the cost surface as complete until both are fixed and covered by mixed-reporting/session-baseline tests.

Review #106
Automated
Reviewed by ChatGPT - Sol

apps/server/src/profileStats.ts
```diff
@@ -897,6 +1154,194 @@ const makeProfileStatsQuery = Effect.gen(function* () {
...
+            ) AS previous_cumulative_cost_usd
+          FROM turns_raw
+        ),
+        turns AS (
+          SELECT
+            turns_with_previous_cost.*,
+            CASE
+              WHEN total_cost_usd IS NOT NULL THEN total_cost_usd
+              WHEN cumulative_cost_usd IS NOT NULL
+                AND previous_cumulative_cost_usd IS NOT NULL
```
kartikkabadi: The first cumulative sample still falls through to `NULL` because `previous_cumulative_cost_usd` is absent. For providers whose adapter emits session-cumulative cost, that first sample is real billable usage and must contribute its full non-negative value; otherwise one-turn sessions become “Not reported” and every longer session is undercounted by its initial baseline.

Review #106 · Automated · Reviewed by ChatGPT - Sol

apps/web/src/components/settings/ProviderUsageActivityCard.tsx
```diff
@@ -0,0 +1,238 @@
...
+      ? "Not reported"
+      : reportedTokens.length === usage.length
+        ? formatCompact(totalTokens)
+        : `${formatCompact(totalTokens)} partial`;
+  const costSummary =
+    reportedCosts.length === 0
+      ? "Not reported"
+      : reportedCosts.length === usage.length
+        ? formatCost(totalCost)
+        : `${formatCost(totalCost)} partial`;
```
kartikkabadi, line 170: A non-null provider cost only proves that *some* turn reported cost, not that every counted turn did. Mixed reporting therefore renders a partial sum as complete. Carry cost-reporting coverage alongside `turnCount` (provider and history as needed) and label incomplete totals as partial/not fully reported instead of treating any non-null sum as complete.

Review #106 · Automated · Reviewed by ChatGPT - Sol

kartikkabadi (commented):
The new machine-activity delta is not ready yet. I independently confirmed the new exact head and the existing unresolved review threads; I am not duplicating the already-anchored inline comments.

Material issues on this head:

- **P1 — usage completeness is still not trustworthy.** The existing provider-wide `providersWithModelUsage` suppression remains, so one model-usage-bearing turn can still hide fallback token history for the same provider. The initial cumulative-cost sample is also still discarded, and mixed cost reporting can still be presented as a complete sum.
- **P2 — machine history can silently become partial.** The OpenCode/Kilo reader caps each database at 50,000 rows but returns an ordinary `ok` activity snapshot with no truncation signal, so high-volume 30-day totals can be understated while looking complete.
- **P2 — Windows default installs are missed.** The database discovery roots do not include the platform-aware `%APPDATA%` location used by the existing OpenCode runtime path resolver, so machine activity can disappear entirely on Windows.
- **P2 — the two usage planes are crossed again.** `enrichWithLocalUsage` appends local machine 24h/7d/30d lines to `snapshot.usageLines` and also attaches the structured `activity`; that makes machine history show inside the account-usage surface and then again in the machine-activity surface, including for providers whose live account usage is explicitly unsupported.
- **P2 — the advertised legacy-second timestamp compatibility is ineffective at the SQL boundary.** The parser converts second-based timestamps to milliseconds, but the query first compares raw `time_created` against a millisecond cutoff, so such rows never reach the parser.

These are all correctness/representation issues, not cosmetic nits. The right end state is to keep account limits and machine activity strictly separate, propagate partial/coverage state whenever bounded readers omit data, reuse the existing platform-aware provider data-directory resolution, and make the Synara-projection accounting coverage-aware rather than inferring completeness from any reported sample.

Exact-head CI for `9ff9e1e2e9526b7782495b7fdbe835d912adc561` is still in progress.

Review #110
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
The two Review #106 cost-accounting findings are still present on this head.

1. The first session-cumulative cost sample is still discarded: the SQL only derives `cost_usd` from `cumulative_cost_usd` when a previous cumulative sample exists. A one-turn cumulative-cost session therefore still appears unreported, and longer sessions remain undercounted by their initial sample.
2. The provider/history cost contract still carries only `costUsd`, with no reporting-coverage signal. If only some completed turns report cost, the partial sum can still be presented as an ordinary total rather than explicitly partial/incomplete.

Exact-head CI is green. I found no new material issue beyond these two unresolved blockers.

Review #107
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
The latest accounting changes fix the earlier first-sample cumulative-cost undercount and add explicit provider-level cost coverage. The turn-boundary token deduplication is also materially better than the old provider-wide suppression.

One correctness issue still remains in the new machine-history path: **Codex and Claude file discovery silently truncates after 2,000 candidate files while the resulting activity snapshot is still reported as complete/`ok`.** On a high-activity installation, valid 30-day history beyond that cap is omitted with no partial-data signal. This is the same completeness class as the earlier bounded SQLite reader issue: either avoid truncating inside the supported window, or propagate truncation into the activity contract/UI so understated totals cannot look complete.

Exact-head CI is also not green: Format and Lint pass, but Typecheck fails, so Test, browser tests, and desktop build are skipped. That must be resolved before this head is considered validated.

Review #111
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Reviewed the new remediation head in two passes. The two blockers from Review #107 are fixed: the first cumulative cost sample now contributes its full value, and provider cost coverage now distinguishes complete, partial, and unreported totals.

One new merge blocker is present on this exact head:

**P1 — the current branch does not typecheck.** Exact-head CI fails in `apps/server` under `exactOptionalPropertyTypes`: `buildMachineUsageActivity()` always passes `partial: input.partial`, so the object contains `partial: undefined` when the caller omits it, while `buildActivityFromMessages()` accepts an optional `partial?: boolean`. The same boundary is deliberately strict elsewhere in this code. Omit the property when it is undefined (or intentionally change the target contract) so the branch compiles. Because Typecheck fails, the main Test, browser-test, and desktop-build steps never run on this head.

I also independently verified two current inline correctness findings remain valid and am not duplicating their inline comments: cumulative token deltas are still computed after telemetry-backed turns are removed from the sequence, which can double-count the filtered turn when `modelUsage` is added back; and cumulative cost `LAG` still partitions only by thread, so a provider/session reset inside one thread can inherit the previous session's baseline.

Exact-head CI: Release Smoke, Migration Lineage, and Windows Process Regression pass; the main quality job fails at Typecheck.

Review #109
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
The Review #111 completeness blocker is fixed on this head. Codex/Claude archive-file truncation and oversized selected archives now propagate a partial machine-activity state, and OpenCode/Kilo database-count truncation is also surfaced instead of presenting bounded totals as complete. The settings UI now preserves that partial detail, including when no usable period can be produced.

I also rechecked the adjacent accounting changes in this delta. I did not find a new material correctness regression worth duplicating as another finding. Existing unresolved review threads outside this delta remain applicable where their code is still current; I am not restating them here.

Exact-head validation is not green: the CI run for `5a6200923ae879e2a37932510dbe92df1d3aedaf` fails at **Format**, so Lint, Typecheck, Test, browser tests, and the desktop build are skipped. This head should not be treated as fully validated until that is fixed and the full pipeline runs.

Review #112
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Reviewed the exact delta from `5a620092` to `28e196f4` in two passes. It is formatting-only: it reformats the SQLite test-table creation and the `resolveClaudeProjectsRoots` signature, with no semantic change. I found no new material issue introduced by this commit.

This head does **not** clear the PR's existing correctness blockers. Current unresolved review findings still include usage-accounting and machine-history correctness issues such as OpenCode/Kilo per-message token totals being treated as cumulative deltas, cumulative-cost baselines crossing filtered/session boundaries, recent-period bucketing problems in Codex/Claude history, and several stale/error/coverage representation issues. I am not duplicating the already-anchored inline comments.

The previous Format failure is the only thing this delta directly addresses. Exact-head CI for `28e196f46ac388af5c6d0527cff554e9e017ec3f` is still in progress, so the branch is not fully validated yet.

Review #113
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
The latest one-line delta is correct as far as it goes: omitting `partial` when it is `undefined` preserves the optional-field contract instead of materializing an explicit undefined value.

It does **not** make this PR clear to merge. The broader branch still has unresolved correctness findings, including incomplete token coverage being presented as complete, database-read failures being collapsed into empty activity, non-indexable provider-history scans, duplicate Claude history across roots, and a macOS launcher path that can lose the app entry point when LaunchServices passes arguments. I am not duplicating those already-anchored inline findings here.

Exact-head CI for `00324c58f065ef71e4d098109a7b00384a084e28` is still in progress.

Review #115
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Exact-head validation follow-up for `00324c58f065ef71e4d098109a7b00384a084e28`: CI is now green. This changes validation status only; it does not clear the existing unresolved correctness findings already anchored on the PR, so I am not posting a duplicate technical review.

Review #116
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Re-reviewed the current head `00324c58f065ef71e4d098109a7b00384a084e28` in two passes: provider/history aggregation first, then bounded machine-history and archive/UI behavior.

The PR is still blocked by material correctness issues:

1. **Cumulative token deltas can be inflated when telemetry-backed turns are filtered before `LAG`.** If a `modelUsage` turn sits between two cumulative `totalProcessedTokens` rows, the later delta is computed against the last retained pre-telemetry row, so it includes the filtered turn and can then be counted again from `modelUsage`. Compute deltas over the complete ordered sequence, then suppress only the telemetry-backed contribution.

2. **Partial machine scans can still render as definitive “no activity.”** The backend now propagates a `partial` status/detail for capped SQLite scans, but the usage panel’s empty-state path treats that result as ordinary absence. A capped scan must remain visibly partial, including when the partial result has no usable periods.

3. **Codex cumulative session totals are bucketed by the final event timestamp rather than by per-event deltas.** A long-running or resumed session spanning the 24-hour/7-day cutoffs can have its older lifetime usage attributed to the newest event and overstated in recent windows.

Exact-head CI run #2518 is green, but these correctness findings remain unresolved.

Review #121
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Re-reviewed the current head `637416e809ebb5c9e4a1f5cdadcf29bcf39ae698` in two passes: provider/accounting aggregation first, then bounded machine-history, archive parity, and UI representation.

The prior Review #121 blockers are addressed on this head:

- cumulative token deltas are computed across the complete ordered sequence before modelUsage-backed turns are zeroed, preventing the filtered-turn double count;
- partial machine scans, including empty-period results, retain an explicit partial state/detail instead of rendering as definitive absence;
- Codex session totals are attributed by per-event deltas within the reporting windows rather than by the final event timestamp.

The related remediation is also covered by focused regression tests for OpenCode/Kilo per-message totals, cumulative-cost baselines across provider/session boundaries, archived-thread parity, archive/database truncation, timestamp-unit handling, and account-versus-machine usage separation.

Exact-head CI run `31335306157` completed successfully; format, lint, typecheck, tests, browser tests, build, and the remaining quality gates are green. I found no new material correctness issue in this delta.

Review #125
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Re-reviewed the new head `eb4009623cd3cfa8ed81fd09dc80445cde2f3b13` in two passes: live profile aggregation first, then deleted-thread archive parity.

**P1 — deleted-thread token aggregation still crosses provider/model boundaries.**

`aggregateThreadTokenRows()` keeps one `previousCumulativeTotal` for the entire thread while computing `totalProcessedTokens` deltas. The live query uses the same thread-wide `LAG`, but its `provider_model_scale` logic and current tests do not establish correctness when a retained thread switches provider/model and the cumulative counter resets or restarts. On the archive path, a sequence such as Codex `100` → Cursor `40` is treated as a reset only when the value decreases; a restart at `140` is interpreted as a `40`-token delta, and a switch back to Codex can then use the Cursor value as its baseline. That can undercount or misattribute archived tokens, while the PR explicitly promises provider/model attribution and archived-thread parity.

The archive mirror needs an explicit session/provider/model boundary key for cumulative-token baselines (or equivalent reset evidence), with regression coverage for provider switches where the next provider's counter is both lower and higher than the previous provider's counter. The live and archive implementations should share the same boundary semantics so deleting a thread cannot change its provider totals.

Exact-head CI status was not changed by this review; the head changed after Review #125, so this is a new technical review rather than a validation-only follow-up.

Review #126
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Exact-head validation follow-up for `eb4009623cd3cfa8ed81fd09dc80445cde2f3b13`: CI run #2558 completed successfully. The current head is green, but Review #126 remains applicable: archived-thread token aggregation can cross provider/model session boundaries and misattribute or undercount cumulative totals after provider switches. No duplicate technical review is being posted; this is a validation-only follow-up.

Review #131
Automated
Reviewed by ChatGPT - Sol

PR #615, opened 2026-08-09
fix(web): isolate ChatView browser tests from cross-test snapshot-sequence leaks

kartikkabadi:
## What

Isolates the ChatView browser tests from cross-test snapshot-sequence leaks. Fixes the intermittent `Browser test (stable)` failure: `ChatView.browser.tsx > ChatView timeline estimator parity (full app) > keeps worktree setup resolvable while attachments upload` (`VitestBrowserElementError: Cannot find element with locator: getByRole('button', { name: 'Cancelling...' })`).

Two mechanisms, both fixed:

1. **Equal-sequence snapshot leak** (`apps/web/src/components/ChatView.browser.tsx`): a late in-flight snapshot with an *equal* sequence number can be accepted over the current test's snapshot, because `isStaleSnapshot` (storeProjection.ts) only drops strictly-lower sequences. Fix: a monotonic snapshot-sequence factory (step 1_000_000) plus a `beforeEach` fixture reset to a neutral sequence-0 shell, so each test starts from a known baseline and a stale equal-sequence snapshot cannot clobber a newer test's state.

2. **Disposed-race snapshot apply** (`apps/web/src/routes/__root.tsx`): a shell snapshot that resolves after transport disposal can still be applied. Fix: `disposed` guards in `loadShellSnapshotOnce` and `DesktopProjectBootstrap` so a snapshot resolving after disposal is never applied.

## Why

The failing test is order/timing dependent: it has failed intermittently on `main`'s own CI (e.g. on 2bd77649) and passed on others. The leak is a harness bug, not a product bug.

## Verification

- `bunx tsc --noEmit` (apps/web) — pass
- `bun run fmt:check` — pass
- Full `ChatView.browser.tsx` browser suite (vitest.browser.stable) — 78 passed, 12 skipped, 0 failed (includes the previously-failing test)

### DIFF
--- a/apps/web/src/components/ChatView.browser.tsx
+++ b/apps/web/src/components/ChatView.browser.tsx
@@ -72,6 +72,17 @@ import { estimateTimelineMessageHeight } from "./timelineHeight";
 
 const THREAD_ID = "thread-browser-test" as ThreadId;
 const OTHER_THREAD_ID = "thread-browser-test-other" as ThreadId;
+
+// Each call to the snapshot factory gets a fresh, monotonically increasing sequence.
+// A generous step between calls guarantees that `recordProjectCreateCommand` and
+// `updateCurrentSnapshot` increments within one test never overlap the next test's
+// base sequence, so a late in-flight shell snapshot from a previous test is always
+// stale and ignored by `isStaleSnapshot`.
+let snapshotSequenceFactory = 0;
+function nextSnapshotSequence(): number {
+  snapshotSequenceFactory += 1000;
+  return snapshotSequenceFactory;
+}
 const THREAD_TITLE = "Browser test thread";
 const UUID_ROUTE_RE = /^\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/;
 const PROJECT_ID = "project-1" as ProjectId;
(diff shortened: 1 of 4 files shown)


kartikkabadi (commented):
## Review summary

I found no material correctness issue in this delta.

The snapshot-sequence isolation is narrowly targeted and addresses the equal-sequence cross-test race without weakening the production stale-snapshot check. Resetting the shared fixture to a neutral sequence before transport disposal, together with the large monotonic test-only sequence step, makes late fixture responses stale rather than authoritative. The new `disposed` guards also correctly prevent post-unmount shell snapshots and repaired snapshots from mutating the read model.

Exact-head CI is still in progress, so the change is not fully validated by the repository pipeline yet. I did not identify a code-level blocker from the diff.

Review #119
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Automated condition-watch follow-up

Exact-head CI for `841aed6942d0c947b8e241685583020e55fb2c60` is now green (`CI` run `31332897175`). This clears the validation uncertainty noted in the prior review.

I found no new material correctness issue in the reviewed delta, and the prior race-hardening assessment remains unchanged. No duplicate technical review was posted for the unchanged code; this is a condition-watch update only.

Review #127
Automated
Reviewed by ChatGPT - Sol

Issue #485, opened 2026-07-28
[Feature]: wsl2 support


jlorezz:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

Not sure

### Problem or use case

not sure if this has been covered but does Synara support wsl2 for windows? i couldn't find a setting for this.

### Proposed solution

-

### Why this matters

-

### Smallest useful scope

-

### Alternatives considered

-

### Contribution

- [ ] I would be open to helping implement this.

cmdr-chara:
I’m working on this in draft PR #744. The first slice owns the WSL process-entry boundary: `\\wsl.localhost\<distro>\...` / `\\wsl$\<distro>\...` workspaces are mapped to Linux paths and launched through `wsl.exe --distribution … --cd … --exec …` without a shell. I’m keeping ACP protocol-level cwd normalization separate so the process and protocol semantics stay independently reviewable.

cmdr-chara:
Closing as completed for the requested WSL2 workspace execution path. Current `main` has the shared WSL boundary in `packages/shared/src/wslBridge.ts` / `platformProcess`, translates `\\wsl.localhost` and `\\wsl$` roots to `wsl.exe --distribution … --cd … --exec …`, normalizes ACP cwd, and includes Windows/WSL regression tests. This shipped in v0.7.3.

PR #744, opened 2026-08-18
feat(windows): run WSL UNC workspaces inside their distro

cmdr-chara:
## What changed

- Recognize WSL2 workspace roots exposed through both `\\wsl.localhost\<distro>\...` and legacy `\\wsl$\<distro>\...` UNC paths.
- Convert those UNC roots into their matching Linux paths.
- Route Windows child-process launches through the system `wsl.exe` using `--distribution`, `--cd`, and `--exec`.
- Keep execution shell-free: command and arguments remain discrete argv entries, so shell metacharacters are not interpreted.
- Skip Windows `where.exe` resolution for WSL workspaces so bare commands resolve from the selected distro's Linux `PATH`.
- Normalize ACP protocol-level workspace paths too: `session/new`, `session/load`, and `session/resume` now receive `/home/...` instead of a Windows WSL UNC path.
- Add focused regression coverage for modern/legacy WSL paths, distro roots, unrelated UNC shares, literal metacharacter handling, and ACP cwd normalization.

## Why

#485 asks for WSL2 support. Synara previously treated Windows workspaces as Win32 execution contexts even when the workspace lived under WSL's UNC filesystem. Entering the distro at process launch is only half of the boundary for ACP providers: the ACP agent must also receive the Linux path in its session request.

The shared Windows process-preparation seam and the shared ACP session runtime now handle those two concerns once, rather than adding WSL-specific logic to each ACP provider.

## Scope

This PR makes UNC-backed WSL workspaces functional through Synara's shared Windows process launcher and ACP runtime.

It deliberately does not add a separate WSL distro picker/execution-mode UI, and it cannot automatically cover a provider-specific launcher that bypasses `prepareWindowsSafeProcess`. Those are separate product/provider concerns rather than reasons to duplicate WSL path conversion here.

Refs #485.

## Validation

Upstream CI is fully green on the current head:

- Format
- Lint
- Typecheck
- Unit tests
- Browser tests (stable + Linux geometry quarantine)
- Desktop build + preload verification
- Windows Process Regression
- Migration Lineage
- Release Smoke

Focused regression coverage lives in:

- `packages/shared/src/windowsWslProcess.test.ts`
- `apps/server/src/provider/acp/AcpWslCwd.test.ts`


cmdr-chara (commented):
Implementation note: this draft intentionally owns only the WSL process-entry boundary. ACP protocol-level cwd normalization is kept out of this diff so the launch and protocol semantics can be reviewed independently.

PR #908, opened 2026-09-02
fix(orchestration): prevent read chats from reappearing as unread after restart

kartikkabadi:
## What this does

Fixes read chats reappearing as unread after a restart. Two root causes, both in the orchestration projection:

1. `projection_threads.updatedAt` could regress: the completion path (`session-set` / `turn-diff-completed`) and the repair-replay path assigned `occurredAt` directly, so a stale completion or a repair replay over a rebuilt projection moved `updatedAt` backwards — and the unread computation reads that timestamp.
2. Existing installs already carry regressed rows: migration `099_InvalidateProjectionThreadsCursor` deletes the `projection.threads` cursor on upgrade so the next bootstrap replays the journal to head and heals every row (precedent: migration 062).

The fix makes all three write paths (`projector.ts`, `Layers/ProjectionPipeline.ts` live + deferred) monotonic via a shared `maxIso` helper hoisted into `turnLifecycle.ts`, and expands the hot-threads event filter so replay covers the completion events.

## Behavior change (user-visible)

- Threads you have read stay read across app restarts and projection rebuilds. No more phantom unread badges after upgrading or after a repair replay.
- One-time cost on first boot after upgrade: the threads projector does a full-journal rescan (batched, crash-safe, idempotent). No data loss, just a slower first boot on very large journals.

## Reviews

- **ponytail ultra: approve.** Minimal root-cause fix at the right rung: shared `maxIso` dedups instead of abstracting, the filter expansion is required for replay coverage, the migration is required for the upgrade path (repair is manual-only). No new dependencies, no new abstractions. Test additions are verbose but each asserts a distinct invariant (rebuild monotonicity, stale-event monotonicity, migration heal).
- **thermos bugs/breakage/security lens: approve, no BLOCKING.** Bootstrap replay verified idempotent and crash-safe; migration SQL fully parameterized; publish behavior unchanged. 5 NITs, all deferred to Follow-ups.
- **thermos maintainability lens: approve, no BLOCKING.** `maxIso` extraction is the right dedup, comments explain the why, tests readable. 8 NITs, all deferred to Follow-ups.
- Blocking findings: none. No fixing commits needed; branch untouched since `641e2c9d`.

## Tests

| Command | Result |
|---|---|
| `bun run test src/persistence/Migrations/099_InvalidateProjectionThreadsCursor.test.ts` (from `apps/server`) | 1 passed |
| `bun run test src/orchestration/projector.test.ts` | 27 passed |
| `bun run test src/orchestration/Layers/ProjectionPipeline.test.ts` | 41 passed |
| `bun run test src/persistence/Migrations.test.ts` | 19 passed |
| `bun run test:browser:stable src/components/SidebarThreadRowContent.browser.tsx` (from `apps/web`) | 2 passed |
| Live headed run: dev stack from this branch, created project + thread in the real UI, full server stop/restart, reloaded in a fresh profile | Thread persists with identical history, zero unread badges; `projection_threads.updated_at` bit-identical pre/post (`2026-09-03T17:44:58.654Z`); threads cursor advanced 38 → 40 by replay; 99 migrations applied |

Live captures (6 screenshots, run video, playwright trace) are local author evidence in the worktree `evidence/908/` dir, not attached here. OS-level screen recording is blocked machine-wide by TCC permissions, so the recording substitute is `run-video.webm` + `trace.zip` (playwright, full action/screenshot timeline).

## Socket surface

None. The diff's only `gateway*` matches are `projection_threads` data columns (`gatewayOperationId`, provenance fields) — no transport, WS, or gateway connect/send/receive logic is touched, so no killed-server failure path applies. The live run above exercised the app's normal WS-driven sync throughout with no connection errors.

## Follow-ups

Deferred NITs from the two thermos passes (no action needed before merge):

- Route the remaining direct `updatedAt` writes (e.g. `turn-start-requested` `createdAt`, deleted/archived paths) through `maxIso`, and add a module note stating which event families need monotonicity (retries/imports/reconciliation can arrive out of order).
- Consolidate the remaining private `maxIso` copy in `Layers/ProjectionSnapshotQuery.ts` onto the canonical `turnLifecycle.maxIso` (verified identical; pre-existing duplication).
- Document the normalized fixed-width ISO-8601 assumption behind lexicographic `maxIso`.
- Migration 099: pin the hardcoded `'projection.threads'` literal to `ORCHESTRATION_PROJECTOR_NAMES.threads` via comment, and note why `thread-shell-summaries` needs no invalidation.
- Tests: why-comment on `DistributiveOmit`, share the triplicated session factory, share the duplicated completion-scenario setup, standardize assert strictness, name the magic cursor count.

## Merge note

- At PR head `641e2c9d`, no rebase (Wave-1 lane). `MERGEABLE`, `CLEAN`, ready, all 9 checks green.
- Safe to merge when approved. Emanuele clicks merge.

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
@@ -82,6 +82,32 @@ const makeAppendAndProject =
       .append(event)
       .pipe(Effect.flatMap((savedEvent) => projectionPipeline.projectEvent(savedEvent)));
 
+type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;
+
+// Scenario event appender: generates the event envelope (ids, causation,
+// metadata) so tests only spell the meaningful fields. `prefix` keeps ids
+// unique across the shared per-file test database.
+const makeScenarioAppender = (
+  appendAndProject: ReturnType<typeof makeAppendAndProject>,
+  prefix: string,
+) => {
+  let scenarioSequence = 0;
+  return (
+    event: DistributiveOmit<
+      Parameters<typeof appendAndProject>[0],
+      "eventId" | "commandId" | "correlationId" | "causationEventId" | "metadata"
+    >,
+  ) =>
+    appendAndProject({
+      ...event,
+      eventId: EventId.makeUnsafe(`evt-${prefix}-${++scenarioSequence}`),
+      commandId: CommandId.makeUnsafe(`cmd-${prefix}-${scenarioSequence}`),
+      correlationId: CommandId.makeUnsafe(`cmd-${prefix}-${scenarioSequence}`),
+      causationEventId: null,
+      metadata: {},
+    });
+};
+
 const exists = (filePath: string) =>
   Effect.gen(function* () {
     const fileSystem = yield* FileSystem.FileSystem;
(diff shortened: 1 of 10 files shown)


kartikkabadi:
Screen recording and frames from the live repro in the PR description. The sidebar shows the Chats section with unread completion dots after restart, which the fix prevents.

01-chats-sidebar

02-composer-focused

03-attach-menu

04-appsnap-added

[Screen recording (.mov)](https://github.com/user-attachments/assets/55856e6c-f356-4c0d-8eb3-1fb3ef374a3d)

kartikkabadi:
Live verification on the fix branch. Kanban board shows the target chat threads in the Chats column, all marked Done, with zero `Unread completion` dots. The screen recording shows the same board loading.

Chats column close-up

Full Kanban board

Initial state

<video controls width="100%" src="https://github.com/user-attachments/assets/3f6fde0f-4e82-4041-8ed7-dbe99eb1bda0"></video>

*Staging media uploaded to kartikkabadi/synara#164.*

PR #920, opened 2026-09-02
feat(web): show all enabled providers in Environment usage section

NachoooLK:
Usage section lists one row per enabled provider again instead of only the active one. Rows without displayable data stay hidden, and the labeled section only mounts once at least one row reports data (keeps a1d5ef156 behavior). Removes the now-unused activeProvider prop from EnvironmentPanel.

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -11377,7 +11377,6 @@ export default function ChatView({
     keybindings,
     availableEditors,
     activeThreadId: activeThread.id,
-    activeProvider: activeThread.session?.provider ?? activeThread.modelSelection.provider,
     isStudioChat: isStudioContainer,
     studioFolderPath: isStudioContainer ? resolvedThreadWorkingDirectory : null,
     showGitActions,
--- a/apps/web/src/components/chat/environment/EnvironmentPanel.tsx
+++ b/apps/web/src/components/chat/environment/EnvironmentPanel.tsx
@@ -14,7 +14,6 @@ import type {
   MessageId,
   PinnedMessage,
   ProjectId,
-  ProviderKind,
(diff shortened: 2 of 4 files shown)


NachoooLK:
@kartikkabadi the initial usage request fan-out flagged by Devin Review (EnvironmentUsageSection.tsx:157-163) is addressed in `8b13b9ada`:

- Rows now defer the provider-scoped `getProviderUsageSnapshot` fallback while the shared batch is in flight. `useProviderUsageSummary` takes a `providerSnapshotPending` input and only enables the provider-scoped query once the batch has settled (success or error). If the settled batch omits a provider, that provider still falls back to its own request.
- Browser coverage added for all three behaviors: no provider-scoped request while the batch is pending, per-provider fallback when the settled batch omits a provider, and full fallback after the batch fails.

Verification:
- `bun run test:browser src/components/chat/environment/EnvironmentUsageSection.browser.tsx`: 7/7 passed
- `bun run test src/components/chat/environment src/lib/serverReactQuery.test.ts`: 57/57 passed
- `bun fmt` / `bun lint` / `bun typecheck` were not run, per the repository's explicit-request rule

NachoooLK:
Rebased `feature/environment-usage-all-providers` onto current `main`. The only conflict was the font-token standardization in `EnvironmentUsageSection.tsx` (`text-[length:var(--app-font-size-chat-meta,10px)]` -> `text-chat-meta`), resolved by keeping the multi-provider rows with the new token.

I also re-ran CI with a no-op `ci: retrigger checks` commit (`596134db8`) because the previous run's failure was the `chat-follow` browser geometry flake (`restores streaming follow after thread switch in the full ChatView`), which also fails on `main` at `b58f27381` and is unrelated to this change. All checks are green now (23/23).

PR #922, opened 2026-09-02
Sidebar visibility pass: persistent status, counted show-more, inline timestamps, live elapsed

NachoooLK:
Scoped sidebar visibility/organization pass (normal + activity views).

- Normal view: status glyph no longer fades under hover (actions overlay with solid bg, no title reflow), Show-more labels with counts, paging keyed by projectId with legacy cwd fallback, short recency on urgent rows, live per-row elapsed ticker reusing the chat Working-for clock (no 0s flash on completion).
- Activity view: inline row timestamps, inline status glyph, no-dim pins, counted Show-more.
- Hover-actions primitive gains optional testId/className (backward compatible).

Verified: web typecheck clean, 186 unit + 8 browser sidebar tests green, fmt/lint clean.

### DIFF
--- a/apps/web/src/components/Sidebar.logic.test.ts
+++ b/apps/web/src/components/Sidebar.logic.test.ts
@@ -27,17 +27,22 @@ import {
   getVisibleThreadsForProject,
   getProjectSortTimestamp,
   hasUnseenCompletion,
+  formatThreadElapsed,
   partitionSidebarThreadsByProjectIds,
   isLatestPinnedThreadMutation,
   isLoopbackHostname,
   isDuplicateProjectCreateError,
   pruneProjectThreadListPagingForCollapsedProjects,
+  pruneProjectThreadListExtraPagesById,
   recoverExistingAddProjectTarget,
   runExclusiveProjectAddition,
   runProjectProvisionWithCancellationRecovery,
   resolvePullRequestReviewBadge,
   resolveSidebarThreadPullRequest,
   resolveThreadDisplayBranch,
+  resolveThreadElapsedMs,
+  resolveUrgentThreadTimeLabel,
+  shouldShowThreadStartingLabel,
   resolveSidebarThreadListPaging,
   resolveProjectEmptyState,
   resolveSettingsBackTarget,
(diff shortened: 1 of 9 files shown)


NachoooLK:
Triage of the Devin review, re-checked against the branch now that it is synced with `main` (merge 315c3fd09 + fix 000607f7d):

**B1 — legacy paging survives project collapse: fixed.**
The cwd-to-id migration now completes as projects resolve (`migrateLegacyProjectThreadListExtraPages`): a legacy cwd value still in use moves into the id map and the cwd entry is dropped. From then on the id-keyed map owns the value, so collapse pruning and "Show less" clearing cannot resurrect a stale cwd entry. Covered in `Sidebar.uiState.test.ts` (migrate, shared cwd, id-wins, unknown cwd untouched).

**B2 — new runs show stale recency: fixed.**
`shouldShowThreadStartingLabel` now compares the session's in-flight `activeTurnId` against the finished turn in the summary: a *different* active turn means a new run has started and its summary has not arrived yet, so the row reads as starting instead of falling through to the previous turn's recency. A matching (or absent) active turn keeps the original no-0s-flash gate on completion.

**B3 — forced rows inflate hidden counts: fixed.**
`deriveSidebarProjectData` now exposes `hiddenRowCount = tree total - rendered rows`, and the chat preview computes hidden as `total - rendered`. The "Show N more (M)" labels count only rows that are genuinely hidden when the active-thread reveal renders past the page cap.

**Flag — activity timestamps do not tick: fixed.**
The Activity view now uses `useNowMs(true, 30s)` for its clock-time/relative labels and date buckets instead of a render-time `Date.now()`.

**Flag — running rows multiply render timers: disagree, keeping per-row scoping.**
Each elapsed label owns one 1s interval that re-renders only its own leaf row, and `startVisibleInterval` stops it while the window is hidden. A shared sidebar clock would need a single tick to re-render the entire (paged) list every second while any row runs; with N concurrent runs the current design is N leaf updates vs. N full-list renders per second. The tradeoff is documented in `SidebarElapsedTimeLabel.tsx`.

The merge also reconciled this branch with the upstream Activity/sidebar refactor (restored the two helpers it deleted, adopted the `text-ui-sm` token). `typecheck`, the full web unit suite (4792 passed) and the sidebar browser tests are green locally.

NachoooLK:
@Emanuele-web04 this one is ready for review: the branch is synced with `main` (conflict resolved), all CI jobs are green, and the Devin Review findings are triaged in a comment above (B1–B3 fixed, the per-row timer flag intentionally kept as-is with rationale). Mergeable/CLEAN.

PR #933, opened 2026-09-03
fix(server): preserve Unicode boundaries in handoff bootstrap

### LINKED ISSUE #932: [Bug]: Handoff transcript truncation can split surrogate pairs and make Codex turn/start time out
## Area

`apps/server`

## Steps to reproduce

1. Create a source thread with at least seven settled user/assistant messages so that at least one imported message is rendered through the 320-code-unit earlier-message summary path.
2. Make an earlier message's normalized text contain a supplementary Unicode character across that truncation boundary. A minimal payload is:

   ```ts
   `${"a".repeat(316)}📌 reminder`
   ```

3. Hand the thread off to Codex.
4. Send the first message in the destination thread.

The current `truncateText` implementation reserves three code units for `...` and calls `slice(0, 317)`. That slice retains only the high surrogate of `📌`.

## Expected behavior

Handoff bootstrap text should stay within the existing character budget while preserving valid Unicode boundaries, and Codex should start the turn normally.

## Actual behavior

The bootstrap contains an unpaired high surrogate. `JSON.stringify` emits it as `\ud83d`, and Codex app-server's Rust JSON parser rejects the complete `turn/start` frame without returning a response. Synara then reports a timeout 20 seconds later.

This reproduced consistently across three handoff attempts. The provider session and Codex thread both started successfully; only `turn/start` failed.

## Impact

Major degradation or frequent failure. The trigger is boundary-dependent, but an affected handoff is deterministically blocked and retrying rebuilds the same invalid bootstrap. Creating a fresh target thread and pasting a summary manually is the current workaround.

## Version or commit

- Synara 0.8.1
- Reproduced against `main` at `562c5fea`

## Environment

- macOS arm64
- Codex CLI 0.153.0
- Provider: Codex
- Source provider in the observed handoff: Claude Agent

## Relevant log

```shell
ERROR codex_app_server_transport::transport: Failed to deserialize JSONRPCMessage: unexpected end of hex escape

Error: Provider adapter request failed (codex) for turn/start: Timed out waiting for turn/start.
```

## Source analysis

`apps/server/src/orchestration/handoff.ts` truncates with a raw UTF-16 offset:

```ts
return `${value.slice(0, Math.max(0, maxChars - 3)).trimEnd()}...`;
```

The repository already contains equivalent surrogate-boundary protection for lossless single-message pagination in `apps/server/src/agentGateway/threadSummary.ts`, but the handoff bootstrap truncator does not use it.

keepure:
## What Changed

- Added shared UTF-16 surrogate-boundary helpers in `@synara/shared/text`.
- Made handoff/bootstrap prefix truncation step back by one code unit when its requested end offset would split a surrogate pair.
- Reused the shared helpers in agent-gateway single-message pagination instead of keeping a duplicate implementation.
- Added focused helper tests and a handoff regression test with `📌` crossing the 320-code-unit earlier-message boundary.

## Why

Fixes #932.

Raw JavaScript slicing can leave an unpaired high surrogate when a supplementary Unicode character crosses the handoff summary boundary. `JSON.stringify` then emits a lone `\ud83d` escape, which Codex app-server rejects without replying to `turn/start`; Synara eventually surfaces only a timeout.

The fix preserves the existing UTF-16 budgets and truncation marker. It changes output only at a split surrogate boundary, where it retains the longest well-formed prefix within the existing budget.

## Verification

- `git diff --check`
- Imported the shared helper directly with Node and confirmed the observed boundary moves from offset 317 to 316, produces a 319-code-unit result, and leaves no unpaired surrogate.
- Sent source-generated failing and corrected JSON-RPC frames to Codex app-server: the lone surrogate reproduces `unexpected end of hex escape`; the complete pair parses normally.

## Checklist

- [x] No UI changes
- [x] No animation or interaction changes

### DIFF
--- a/apps/server/src/agentGateway/threadSummary.ts
+++ b/apps/server/src/agentGateway/threadSummary.ts
@@ -13,6 +13,7 @@ import type {
   OrchestrationThread,
   OrchestrationThreadShell,
 } from "@synara/contracts";
+import { splitsSurrogatePair, unicodeSafeEndOffset } from "@synara/shared/text";
 
 export type AgentThreadStatus =
   | "working"
@@ -219,24 +220,6 @@ function readMessageCoordinates(input: {
     messageId: input.messageId,
     messageVersion: input.messageVersion,
   };
-}
-
-function splitsSurrogatePair(text: string, offsetChars: number): boolean {
-  if (offsetChars <= 0 || offsetChars >= text.length) return false;
-  const previousCodeUnit = text.charCodeAt(offsetChars - 1);
-  const nextCodeUnit = text.charCodeAt(offsetChars);
-  return (
-    previousCodeUnit >= 0xd800 &&
-    previousCodeUnit <= 0xdbff &&
-    nextCodeUnit >= 0xdc00 &&
-    nextCodeUnit <= 0xdfff
-  );
-}
-
-function unicodeSafeEndOffset(text: string, requestedEndOffsetChars: number): number {
-  return splitsSurrogatePair(text, requestedEndOffsetChars)
-    ? requestedEndOffsetChars - 1
-    : requestedEndOffsetChars;
 }
 
 function paginateSingleMessage(input: {
--- a/apps/server/src/orchestration/handoff.test.ts
+++ b/apps/server/src/orchestration/handoff.test.ts
@@ -3,10 +3,10 @@
 // Layer: Orchestration mapping tests
 // Depends on: handoff.
 
-import { MessageId, type OrchestrationMessage } from "@synara/contracts";
+import { MessageId, type OrchestrationMessage, ThreadId } from "@synara/contracts";
 import { describe, expect, it } from "vitest";
 
-import { buildPriorTranscriptBootstrapText } from "./handoff.ts";
+import { buildHandoffBootstrapText, buildPriorTranscriptBootstrapText } from "./handoff.ts";
 
 const message = (
   index: number,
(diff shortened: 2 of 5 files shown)


Emanuele-web04 (commented):
Reviewed against #932. The fix is correct and mergeable; the notes below are non-blocking.

**Verified locally**
- `bun fmt`, `bun lint` (0 warnings on touched files), `bun typecheck` all pass.
- `handoff.test.ts`, `threadSummary.test.ts`, `text.test.ts` pass (23 tests).
- Reverting `handoff.ts` to the pre-fix version makes the new regression test fail, so it genuinely guards the bug.
- The issue payload (`"a".repeat(316) + "📌 reminder"`) now yields a 319-code-unit, well-formed result; `JSON.stringify` no longer emits a lone `\ud83d`.

**Correctness**
- `truncateText` (`apps/server/src/orchestration/handoff.ts:18-25`) is the only raw-slice site in that file and covers all three callers (recent messages at 2,400, earlier summaries at 320, and the final 32,000 ceiling), so every handoff/fork/prior-transcript bootstrap path is fixed.
- Budget is still respected: at most one code unit is dropped, and the `maxChars < 4` behavior is unchanged from before (`"..."` for 0..3).
- Moving `splitsSurrogatePair`/`unicodeSafeEndOffset` verbatim into `packages/shared/src/text.ts` and reusing them from `threadSummary.ts` is exactly what CLAUDE.md asks for instead of a second copy. Nice.

**Nits / follow-ups (optional, could be separate PRs)**
1. `apps/server/src/agentGateway/threadSummary.ts:131` and `:304` still truncate with raw `text.slice(0, n)` in the same file the helper was lifted from. Now that `unicodeSafeEndOffset` is shared, those two could use it in a one-line change each.
2. The same class of bug exists in other provider-bound prompt text, e.g. `apps/server/src/provider/threadMentionContext.ts:59` (`block.slice(0, availablePrefixChars)`) and `apps/server/src/orchestration/providerRuntimeActivityProjection.ts:109`. Out of scope for #932, but worth a follow-up sweep with the new helper.
3. The helper only protects surrogate pairs, not grapheme clusters (variation selectors, ZWJ sequences, combining marks). Splitting those still produces valid UTF-16/JSON, so it can't reproduce the Codex parse failure; just noting the documented scope is intentionally narrow.
4. Defense in depth: a lone surrogate already present in source message text would still reach Codex. Applying `String.prototype.toWellFormed()` at the JSON-RPC transport boundary would make that class of failure impossible regardless of any truncator. Separate change, not needed here.
5. Test nit: `handoff.test.ts` could also assert the summary line length is `<= 320` to lock the budget invariant, though `toContain(\`${"a".repeat(316)}...\`)` implies it.

PR #1016, opened 2026-09-07
fix(models): normalize provider model display names

SHLE1:
## What Changed

- Fix catalog name precedence so generated custom and selected-model entries no longer override provider-supplied display names.
- Replace broad identifier-equivalence matching with GLM/DeepSeek-specific name normalization, preserving variant wording such as `Highspeed` and other brands’ casing such as `MiniMax M2.5 Free` and `K2P6`.
- Restore brand casing in `humanizeModelSlug` via a small token map (`glm` -> `GLM`, `deepseek` -> `DeepSeek`).
- Rename static contracts entries `GLM-5.2` / `GLM-5.2 Fast` / `GLM-5.1` to space style (`GLM 5.2`, etc.) for consistency.

## Why

Pi provides display names such as `GLM-5.3-Flash`. Selecting a model that is absent from the static catalog adds a generated fallback entry to the local options. During catalog merging, that generated name previously took precedence over the discovered name, changing `GLM-5.3-Flash` to `Glm 5.3 Flash` after selection.

The fix prevents generated entries from overriding discovered names and applies explicit GLM/DeepSeek formatting rules. This keeps labels stable while preserving provider-supplied branding for other models, including MiniMax and K2P6. The shared catalog merge correction applies to both Pi and OpenCode, as well as other providers using this path.

## UI Changes

Text-only label change in the model picker, no layout change:

- Before: `Glm 5.3 Flash`, `Deepseek V4 Flash` (mixed brand casing, label style changes after clicking)
- After: `GLM 5.3 Flash`, `DeepSeek V4 Flash` (consistent brand casing, label stays stable after selection)
- Preserve variant wording such as `Highspeed` and branded names such as `MiniMax M2.5 Free` and `K2P6`.

| Before | After |
| --- | --- |
| Before - Pi model picker showing old labels | After - Pi model picker showing normalized labels |

## Verification

- From `apps/web`: `bun run test -- src/providerModelOptions.test.ts src/appSettings.test.ts` — 102 tests passed, including selected-model placeholder regression coverage for Pi and OpenCode.
- From `packages/shared`: `bun run test -- src/model.test.ts` — 103 tests passed.
- From `apps/web`: `bun run test:browser -- src/components/chat/ProviderModelPicker.browser.tsx` — 18 tests passed.
- Changed-file formatting checks passed, including the three files flagged by CI.
- Full-workspace lint and typecheck have not been rerun locally.

## Checklist

- [ ] I included a video for animation/interaction changes

### DIFF
--- a/apps/web/src/providerModelOptions.ts
+++ b/apps/web/src/providerModelOptions.ts
@@ -46,6 +46,12 @@ export interface ProviderModelOptionGroup {
   options: ProviderModelOption[];
 }
 
+// Catalog names that differ from their model id only by separators or casing
+// are still id-derived labels rather than intentional presentation metadata.
+function modelDisplayIdentity(value: string): string {
+  return value.trim().toLowerCase().replace(/[-_\s]+/gu, "");
+}
+
 /**
  * Returns the provider provenance shown when a model is detached from its
  * normal upstream-provider group (for example, inside Favourites).
@@ -162,6 +168,11 @@ export function mergeDynamicModelOptions(input: {
 
     const normalizedSlug = normalizeDynamicModelSlug(input.provider, dynamicModel.slug);
     const rawSlug = dynamicModel.slug.trim().toLowerCase();
+    const modelIdentifier = normalizedSlug.includes("/")
+      ? normalizedSlug.slice(normalizedSlug.lastIndexOf("/") + 1)
+      : normalizedSlug;
+    const rawNameIsIdentifier =
+      rawName.length > 0 && modelDisplayIdentity(rawName) === modelDisplayIdentity(modelIdentifier);
     const displayNameFallback = formatProviderModelOptionName({
       provider: input.provider,
       slug: normalizedSlug,
@@ -176,7 +187,8 @@ export function mergeDynamicModelOptions(input: {
         staticNameBySlug.get(normalizedSlug) ??
         (rawName.length > 0 &&
         rawName.toLowerCase() !== rawSlug &&
-        rawName.toLowerCase() !== normalizedSlug.toLowerCase()
+        rawName.toLowerCase() !== normalizedSlug.toLowerCase() &&
+        !rawNameIsIdentifier
           ? rawName
           : displayNameFallback),
       ...(dynamicModel.description?.trim() ? { description: dynamicModel.description.trim() } : {}),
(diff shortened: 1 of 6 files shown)


apps/web/src/providerModelOptions.ts
```diff
@@ -162,6 +168,11 @@ export function mergeDynamicModelOptions(input: {
 
     const normalizedSlug = normalizeDynamicModelSlug(input.provider, dynamicModel.slug);
     const rawSlug = dynamicModel.slug.trim().toLowerCase();
+    const modelIdentifier = normalizedSlug.includes("/")
+      ? normalizedSlug.slice(normalizedSlug.lastIndexOf("/") + 1)
+      : normalizedSlug;
+    const rawNameIsIdentifier =
+      rawName.length > 0 && modelDisplayIdentity(rawName) === modelDisplayIdentity(modelIdentifier);
```
kartikkabadi: Thanks for the PR. The `glm` and `deepseek` casing fixes are right, but `rawNameIsIdentifier` looks like it will regress provider-branded names that aren't in `MODEL_TOKEN_DISPLAY_NAMES`.

OpenCode serves `opencode/minimax-m2.5-free` with `name: "MiniMax M2.5 Free"`. With this check, `modelDisplayIdentity("MiniMax M2.5 Free")` equals `modelDisplayIdentity("minimax-m2.5-free")`, so the provider-supplied name is discarded and the fallback is `humanizeModelSlug`. That function only restores casing for `deepseek` and `glm`, so `minimax` becomes `Minimax` and the rendered label is `Minimax M2.5 Free` — losing the brand casing.

Same issue clobbers `kimi-for-coding/k2p6` with `name: "K2P6"` → `K2p6`.

Also, `Static Checks (fast)` is failing on the `Format` step (`bun run fmt:check` / `oxfmt --check`). It flags formatting in 3 files: `apps/web/src/components/chat/ProviderModelPicker.browser.tsx`, `apps/web/src/providerModelOptions.ts`, and `packages/shared/src/model.ts`. Running `bun run fmt` will fix the import/chain indentation.

Could we keep the provider's exact casing when the raw name is only a casing/separator variant, or expand the token table to cover `minimax`, `kimi`, and other provider brands discovery is known to return?

SHLE1: Good catch! 

you're right, I am trying to find a better path, and have already discovered something deeper. 

Will update later.

PR #1122, opened 2026-09-10
Bound provider start checkpoint latency

iiiMohammed:
## Summary

- bound the advisory message-start checkpoint to 5 seconds across Git mutation queueing and capture work, while allowing the provider turn to continue on budget exhaustion
- cap `git add` output, preflight oversized unseeded workspaces, and suppress repeated expensive attempts for 60 seconds
- serialize checkpoint mutations through `GitCore.withMutation` and return a typed, path-safe budget error
- preserve the existing 180-second behavior for authoritative checkpoint flows that do not opt into the advisory policy

PR #1115 fixed orchestration event publication backpressure and the visible pre-start timeout path. This change addresses the remaining pre-provider checkpoint work that can still delay `thread.turn.start` in large or noisy Windows workspaces.

## Testing

- `bun run --cwd apps/server test -- src/checkpointing/Layers/CheckpointStore.test.ts src/orchestration/Layers/ProviderCommandReactor.test.ts src/orchestration/Layers/OrchestrationEngine.test.ts` — 260 passed
- post-format `CheckpointStore.test.ts` — 9 passed
- `bun run typecheck` — 7/7 packages passed
- focused `oxfmt --check` — passed
- focused `oxlint` — 0 errors (4 existing warnings)
- `git diff --check` — passed

Broader Windows runs still hit unrelated baseline failures. On the unchanged `e307d251` base, the same two `codexAppServerManager` symlink/auth tests fail, and representative `GitCore` path/performance tests fail with the same results. The root suite's two POSIX-path canary failures and the default-timeout `CheckpointReactor` case were also reproduced on unchanged base; the representative reactor case passes with a 30-second test timeout.

### DIFF
--- a/apps/server/src/checkpointing/Errors.ts
+++ b/apps/server/src/checkpointing/Errors.ts
@@ -35,8 +35,29 @@ export class CheckpointInvariantError extends Schema.TaggedErrorClass<Checkpoint
   }
 }
 
+/**
+ * CheckpointCaptureBudgetExceededError - Advisory capture exceeded its resource policy.
+ *
+ * Intentionally excludes workspace paths and subprocess output so a skipped
+ * pre-turn checkpoint can be logged without exposing filenames.
+ */
+export class CheckpointCaptureBudgetExceededError extends Schema.TaggedErrorClass<CheckpointCaptureBudgetExceededError>()(
+  "CheckpointCaptureBudgetExceededError",
+  {
+    operation: Schema.String,
+    reason: Schema.Literals(["timeout", "unseeded-scan", "cooldown"]),
+    timeoutMs: Schema.Number,
+    detail: Schema.String,
+  },
+) {
+  override get message(): string {
+    return `Checkpoint capture skipped (${this.reason}): ${this.detail}`;
+  }
+}
+
 export type CheckpointStoreError =
   | GitCommandError
+  | CheckpointCaptureBudgetExceededError
   | CheckpointInvariantError
   | CheckpointUnavailableError;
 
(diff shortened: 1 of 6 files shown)


iiiMohammed:
@kartikkabadi thanks — both bugs are blocking findings, and I’ve included both flags in the repair scope.

The resolution must prevent an advisory owner’s short timeout from aborting an authoritative waiter, preserve the advisory cooldown across an authoritative success unless a policy-bearing advisory capture succeeds, key suppression consistently for linked worktrees, and keep unrelated Git failures distinguishable from budget exhaustion.

I’ll verify this with targeted owner/waiter race tests, cooldown timing tests, linked-worktree coverage, and error-classification assertions before reporting it resolved.

iiiMohammed:
﻿Pushed follow-up commit `33d05cdb4` after merging current `main`.

This addresses the two review bugs and the two related flags:

- Advisory and authoritative captures now use separate in-flight slots, so an advisory caller deadline cannot interrupt an authoritative waiter.
- An authoritative success no longer clears an active advisory failure cooldown.
- Advisory cooldowns are keyed by the repository's canonical Git common directory, so linked worktrees share the cooldown.
- The unseeded preflight maps only Git timeout/output-overflow failures to `CheckpointCaptureBudgetExceededError`; unrelated Git failures remain `GitCommandError`.

The new regression tests failed in all four cases before the implementation change and pass afterward.

Validation:

- `CheckpointStore.test.ts`: 13/13 passed
- `ProviderCommandReactor.test.ts`: 238/238 passed
- Combined focused run: 251/251 passed
- `apps/server` typecheck passed
- Focused `oxfmt --check` and `oxlint` passed

Windows reproduction on the currently installed 0.8.4 build showed provider-start checkpoint delays of 34.39s and 34.35s, while the subsequent DeepSeek responses took 1.13s and 1.23s. That evidence attributes the visible delay to checkpoint capture over the broad project root, not DeepSeek inference. This PR updates source behavior only; the installed build was not modified during validation.

PR #1137, opened 2026-09-11
Unify file manager labels across platforms

iiiMohammed:
## Summary

- centralize platform- and target-specific file-manager labels and error copy in a pure web helper
- use a thin toast adapter across the project menu, file references, editor metadata, and Studio environment surfaces
- preserve `showInFolder` while explicitly testing folders through `shell.openPath` and files through `shell.showItemInFolder`
- keep server file-manager launch behavior unchanged

## Focused verification

- web naming, toast, file-reference, and editor metadata: 41 passed
- Sidebar and Environment guards: 125 passed
- desktop file/folder dispatch: 12 passed
- server file-manager case: 1 passed, 23 skipped by the focused filter
- `git show --check`: clean

The full Windows run of `apps/server/src/open.test.ts` produced 22 passes and two unrelated baseline failures: Linux terminal selection expected `konsole` but found `x-terminal-emulator`, and Windows Store VS Code discovery returned false. Neither `apps/server/src/open.ts` nor `apps/server/src/open.test.ts` is changed by this PR.

## Verification not run

Per the task constraints, `bun fmt`, `bun lint`, and `bun typecheck` were not run. Native file-manager UI smoke checks on macOS/Linux also remain outside this Windows-only environment.

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -155,6 +155,7 @@ import {
 } from "./resumableUpdateDownload";
 import { hardenElectronUpdater } from "./electronUpdaterSecurity";
 import { ServerListeningDetector } from "./serverListeningDetector";
+import { showPathInFileManager } from "./showInFileManager";
 import { BackendStartupBlockDetector, type BackendStartupBlock } from "./backendStartupBlock";
 import {
   BACKEND_MAX_CONSECUTIVE_START_FAILURES,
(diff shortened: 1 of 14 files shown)


iiiMohammed:
@kartikkabadi thanks — I’ve added this as a required compatibility check.

I’ll verify the pre-change handler behavior and Electron `shell.openPath` contract directly, then make the implementation and test agree with the established semantics. I won’t treat the whitespace case as resolved—or change it merely to satisfy the new test—until that evidence is pinned and the focused regression passes.

Issue #490, opened 2026-07-30
Antigravity: inactive synara-capture PreToolUse hook returns invalid empty object and blocks every tool


3mrr4dy:
## Summary

The Antigravity capture plugin installs a global `PreToolUse` hook with `matcher: "*"`. When `SYNARA_ANTIGRAVITY_EVENTS` is not set, its fallback writes `{}` to stdout.

Antigravity requires `PreToolUse` to return a `decision` (`allow`, `ask`, `deny`, or `force_ask`). Because the fallback omits it, every tool call is rejected before execution with an empty denial reason.

## Reproduction

1. Use Synara's Antigravity integration with the generated `synara-capture/hooks.json` installed globally.
2. Ensure `SYNARA_ANTIGRAVITY_EVENTS` is unset for the Antigravity language-server process.
3. Start or restart Antigravity.
4. Invoke any tool: `list_dir`, `run_command`, `view_file`, `search_web`, `ask_permission`, etc.

## Actual result

Every tool fails immediately:

```
Error invalid tool call: model output error: invalid tool call error (invalid_args) tool call denied with reason:
```

The reason is empty. This reproduces across conversations and tool types because the hook is global and matches all tools.

## Expected result

When capture is inactive, the hook should preserve normal Antigravity permission behavior instead of invalidating every tool call.

## Root cause

The generated fallback in `hooks.json` is effectively:

```sh
if [ -z "${SYNARA_ANTIGRAVITY_EVENTS:-}" ]; then
  cat >/dev/null 2>&1 || :
  printf '%s\n' '{}'
fi
```

For **PreToolUse**, that output is invalid because `decision` is required.

## Proposed fix

Change only the inactive `PreToolUse` fallback to:

```json
{"decision":"ask"}
```

For example:

```sh
if [ -z "${SYNARA_ANTIGRAVITY_EVENTS:-}" ]; then
  cat >/dev/null 2>&1 || :
  printf '%s\n' '{"decision":"ask"}'
fi
```

Keep `{}` for `PostToolUse`, `PreInvocation`, and `PostInvocation`, where it is valid.

## Environment

- macOS 26.5
- Synara 0.6.3
- Antigravity IDE 2.4.3
- agy CLI 1.1.8

I applied the one-line local workaround to both the IDE-discovered and CLI plugin copies, restarted Antigravity, and verified that the pre-tool hook now returns `{"decision":"ask"}`.

Issue #465, opened 2026-07-26
[Bug]: Antigravity: thread stays "working" after assistant finishes; Cancel does nothing; follow-ups blocked


Kjell6:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Open Synara Desktop with Antigravity enabled and authenticated (`agy` CLI available).
2. Create a new thread (or open an existing one) with provider **Antigravity** and any available model (e.g. Gemini 3.6 Flash).
3. Send a user message.
4. Wait until the assistant has fully finished and the complete reply is visible in chat.
5. Observe the thread status indicator and the composer.
6. Click **Cancel** while the status still shows Working (during or after the reply is already complete).
7. Try to send a follow-up message.

### Expected behavior

-  When the assistant turn finishes, the thread returns to **Idle** (or equivalent non-working state).
- The composer immediately accepts a new message.
- **Cancel** stops the in-flight Antigravity turn and unlocks the composer promptly.

### Actual behavior

- **Cancel does nothing** (or has no effect on Antigravity turns).
- After the model has **already finished and rendered a full reply**, the thread still shows **Working**.
- The composer stays blocked; **no follow-up messages** can be sent.
- The stuck Working state can persist for a long time. Recovery sometimes only happens after a long wait, closing the app or machine sleep-wake .
- Other providers do not show this; it appears specific to the Antigravity integration.

### Impact

Blocks work completely

### Version or commit

Version 0.6.1 (3f66f8ee5b1e)

### Environment

macos 26.5.2 

### Logs or stack traces

```shell

```

### Workaround

- use another provider
- App restart

aristotl-dylan:
This looks fixed on current `main`: #542 (`44a32d11a`, *settle Antigravity turns on cancel and clean stop*) plus the follow-up #552 (`f47959f4d`), both shipped in **v0.7.1**. The adapter now emits a single idempotent terminal `turn.completed` from process-close, interrupt, and stop paths (the settle path in `AntigravityAdapter.ts` references this issue directly).

The report was against v0.6.1 — worth re-testing on v0.7.1+, but this should be safe to close.

cmdr-chara:
[...]

PR #1143, opened 2026-09-11
fix(antigravity): stop sending a decision from the PreInvocation hook

iiiMohammed:
## The bug

Antigravity turns fail with **"Antigravity CLI exited with code 1."** even though the assistant reply has already streamed into the thread. The turn is marked failed, the thread goes to `error`, and `lastError` is that message.

## Root cause

`inactiveHookOutput()` answered PreInvocation with `{"decision":"allow"}`. The Antigravity CLI decodes PreInvocation hook output with **protojson**, into a message that has **no `decision` field**, so the answer never unmarshals:

```
E prehooks.go:43] failed to call custom pre-invocation hook
  jsonhook__synara-capture_PreInvocation_0_0: failed to unmarshal result from hook
  jsonhook__synara-capture_PreInvocation_0_0 via protojson:
  {"decision":"allow"}: proto: (line 1:2): unknown field "decision"
```

A failed PreInvocation aborts the invocation it gates, and the parent CLI exits 1. `AntigravityAdapter` then settles the turn as `failed` with `Antigravity CLI exited with code ${code}.`, which is what the user sees.

The current comment in the file states the opposite — that `{}` is treated as a denial and `"allow"` is required. That holds for PreToolUse (#490), but not for PreInvocation.

## Evidence

From `~/.gemini/antigravity-cli/log/` on Windows, Antigravity CLI `1.2.1`:

- Every failing Synara-managed run hit this error, and **only** on `PreInvocation` — 10/10 failures were `jsonhook__synara-capture_PreInvocation_0_0`, never `PreToolUse`.
- Correlating a failed turn: assistant message delivered at `15:26:38Z`, `turn.completed / Turn failed` at `15:26:40Z` — the crash is at exit, after the reply.
- Locally forcing the pre-invocation answer to `{}` produced **zero** `prehooks.go` errors across 25+ subsequent CLI runs, and Antigravity turns completed normally.

There are two distinct failure texts depending on how the answer reaches the CLI, both from the same cause:

- `proto: (line 1:2): unknown field "decision"` — well-formed JSON, rejected field.
- `proto: syntax error (line 1:2): invalid value \` — the win32 fallback, where JSON escapes survive into `cmd` (already documented in `buildAntigravityCaptureCommand`) so the `"` arrive as `\"`. `{}` has no quotes, so this path stops mattering for PreInvocation too.

## The change

- `inactiveHookOutput()` no longer special-cases `pre-invocation`; it falls through to `{}`.
- `hookScriptSource()` answers `{}` for pre-invocation in both the inactive fallback and the active branch, and still records the event.
- PreToolUse is untouched: it does carry a decision, and `{}` there denies every tool call (#490).
- Comments updated — the previous ones asserted the inverted semantics.

## Tests

- Updated the three existing PreInvocation expectations (inactive fallback, win32 command, darwin command).
- Added `answers pre-invocation without a decision while still capturing the event`: asserts the emitted object has no keys even when `SYNARA_ANTIGRAVITY_HOOK_DECISION=allow` is set, and that the NDJSON event is still written. It spawns the capture script directly, because the win32 wrapper is deliberately quote-free and cannot launch a helper whose path contains a space.

## Verification

- `oxfmt --check` clean on both files; `oxlint` reports 0 errors.
- `AntigravityAdapter.test.ts`: 42 passed, 2 failed — the **same 2** that fail on unmodified `main` in this environment (`runs the capture script for Synara-managed sessions` and the #465 NDJSON test). Both spawn `process.execPath`, which here is `C:\Program Files\nodejs\node.exe`; the space breaks the intentionally quote-free win32 hook command. No new failures.
- Full `typecheck` was not validated locally: the worktree borrows `node_modules` from a v0.8.3 checkout, which produces 107 pre-existing errors in unrelated files. **0** of them are in the two changed files.

### DIFF
--- a/apps/desktop/src/appSnapManager.test.ts
+++ b/apps/desktop/src/appSnapManager.test.ts
@@ -1020,12 +1020,28 @@ describe("AppSnap window picker requests", () => {
 
   it("captures a specific window and keeps a pending record until acknowledged", async () => {
     const { manager, watchChild, captureDirectory, dispose } = await createEnabledManager();
+    const capturePath = join(captureDirectory, "req-capture.png");
+    const originalUnlink = FS.promises.unlink;
+    let markHelperUnlinkStarted!: () => void;
+    const helperUnlinkStarted = new Promise<void>((resolve) => {
+      markHelperUnlinkStarted = resolve;
+    });
+    let releaseHelperUnlink!: () => void;
+    const helperUnlinkRelease = new Promise<void>((resolve) => {
+      releaseHelperUnlink = resolve;
+    });
+    const unlinkSpy = vi.spyOn(FS.promises, "unlink").mockImplementation(async (path) => {
+      if (path.toString() === capturePath) {
+        markHelperUnlinkStarted();
+        await helperUnlinkRelease;
+      }
+      return originalUnlink(path);
+    });
     try {
       const capturing = manager.captureWindow(77);
       await flushPromises();
       const requestId = lastStdinLine(watchChild).slice("capture-window ".length, -3);
       expect(requestId).toMatch(/^picker-[a-f0-9-]+$/);
-      const capturePath = join(captureDirectory, "req-capture.png");
       const captureBytes = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 7]);
       writeFileSync(capturePath, captureBytes);
       watchChild.stdout.write(`${JSON.stringify({ type: "triggered", id: requestId })}\n`);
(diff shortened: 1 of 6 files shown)


iiiMohammed:
@kartikkabadi thanks — I’m treating the cleanup hang as a blocking finding and the two scope/behavior flags as required investigations.

The repair will keep a claimed capture cancelable through cleanup or otherwise bound cleanup by the request lifecycle, so disable/teardown can still settle the renderer promise. I’ll also verify whether the AppSnap and transcript-scroll changes should be separated or documented, and test keyboard reflow against the wheel-path behavior rather than assuming symmetry.

The acceptance coverage will include a stalled `unlink`, helper termination/disable while cleanup is pending, and keyboard reflow compensation.

PR #1147, opened 2026-09-12
Fix live turn ownership during runtime reconciliation

iiiMohammed:
## What Changed

Preserve a live provider turn's ownership when its projected lifecycle is older than 45 minutes. Reconciliation can realign an incorrectly interrupted manual turn, including clearing its old completion timestamp, while preserving completed/error turns.

Include ownerless `starting`/`running` sessions in reconciliation candidates and report `provider_lifecycle_stale` from thread diagnosis when the recovery grace period has elapsed.

## Why

Thread/session timestamps describe shell and lifecycle changes; assistant text and tools do not refresh them. Previously, the 45-minute fallback could mark a still-running Codex turn interrupted and clear its binding. A scheduled continuation then passed the projection-based eligibility check and left the thread `starting` with an interrupted latest turn. The provider reactor correctly queued against the live turn, but the old caller had already lost MCP write authority. Candidate selection excluded the resulting state and diagnosis had no corresponding finding.

The fix retains exact active-turn authorization and existing queue/steer semantics. Truly stalled providers remain the responsibility of their runtime watchdog/stop path, which must stop the actual owner before settling it.

## Verification

- Regression tests first reproduced the false interruption, omitted startup candidate, and empty diagnosis on the original implementation.
- Windows process integration uses the production Codex adapter, process manager, durable orchestration/projection/queue, and MCP authorization against a deterministic local app-server protocol process. Both healthy long-turn and legacy interrupted/starting cases complete manual and automation turns in order. Synara, Git, SQLite, and Codex homes are isolated; the subprocess guard rejects accidental real Codex launches. This is protocol acceptance, not a hosted-model test.
- Dedicated automation coverage exercises scheduler deferral, the false-interruption eligibility gap, exact run/turn association, and silent completion reporting.
- Focused regression suites: 341 passing tests, including both new process integration cases.
- The process fixture also emits the observed Codex model-refresh stderr warning during initialization. Both integration cases pass with the warning projected, the first turn starting without retry, and `lastError` remaining null. Server typecheck and formatting/lint for the two updated test files pass.
- Workspace typecheck: 7/7 packages. Lint: zero errors (559 warnings).
- Formatting: all 3,181 matched files pass on an LF export of the staged Git tree; the Windows checkout has pre-existing CRLF and unrelated untracked planning files.
- Broad server sweep (provider/orchestration/automation/gateway and existing integrations, excluding ProviderHealth): 2,778 passed, 34 failed, 5 skipped across 192 files. Seven failures were CRLF expectations; the eight selected checkpoint tests and the affected integration test all passed with process-local `core.autocrlf=false`. All remaining **27/27 failures reproduced on unchanged main** (`a355cf2000b18aeb032e7dac18b74efdb3ab3268`) in an isolated Git archive with the same locked dependencies. These are existing platform/path, hook, discovery, SDK-fixture, and atomic-file tests; no changes were made to those implementations.
- The initial broad run stalled in ProviderHealth after its update-timeout assertion failed. That failure also reproduces on unchanged main; that file was excluded from the completed sweep. The repository-wide Windows suite is therefore not green, while the lifecycle change and its focused/integration checks are verified.
- [CI on the implementation commit](https://github.com/Emanuele-web04/synara/actions/runs/34664474540) passed both server shards, Windows process regression, static checks, and the desktop build. It failed the AppSnap temporary-file assertion and ChatView wheel-follow geometry assertion. The same ChatView assertion fails on [unchanged main](https://github.com/Emanuele-web04/synara/actions/runs/34646617440). The focused AppSnap test passes locally on both base and head; its existing capture promise settles before asynchronous temporary-file deletion. Desktop, web, shared contracts, and locked dependencies are unchanged by this PR. The latest test-only commit has a fresh CI run.

## Checklist

- [x] This PR is small and focused: lifecycle repair and regression coverage.
- [x] I explained what changed and why.
- [x] UI screenshots and animation video are not applicable to this server change.

### DIFF
--- a/apps/server/integration/OrchestrationEngineHarness.integration.ts
+++ b/apps/server/integration/OrchestrationEngineHarness.integration.ts
@@ -47,6 +47,9 @@ import { makeProviderServiceLive } from "../src/provider/Layers/ProviderService.
 import { makeCodexAdapterLive } from "../src/provider/Layers/CodexAdapter.ts";
 import { CodexAdapter } from "../src/provider/Services/CodexAdapter.ts";
 import { ProviderService } from "../src/provider/Services/ProviderService.ts";
+import { ProviderRuntimeReconciler } from "../src/provider/Services/ProviderRuntimeReconciler.ts";
+import { makeProviderRuntimeReconcilerLive } from "../src/provider/Layers/ProviderRuntimeReconciler.ts";
+import { ProviderRuntimeEventRepositoryLive } from "../src/persistence/Layers/ProviderRuntimeEvents.ts";
 import { ServerSettingsService } from "../src/serverSettings.ts";
 import { CheckpointReactorLive } from "../src/orchestration/Layers/CheckpointReactor.ts";
 import { StudioOutputReactorLive } from "../src/orchestration/Layers/StudioOutputReactor.ts";
@@ -176,6 +179,7 @@ export interface OrchestrationIntegrationHarness {
   readonly engine: OrchestrationEngineShape;
   readonly snapshotQuery: ProjectionSnapshotQuery["Service"];
   readonly providerService: ProviderService["Service"];
+  readonly runtimeReconciler: ProviderRuntimeReconciler["Service"];
   readonly checkpointStore: CheckpointStore["Service"];
   readonly checkpointRepository: ProjectionCheckpointRepository["Service"];
   readonly waitForThread: (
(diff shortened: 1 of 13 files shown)


iiiMohammed:
@kartikkabadi thanks — the starvation example is a valid blocking scenario.

The repair will make reconciliation advance past matching live/no-plan candidates using a stable bounded cursor or bounded overfetch, while preserving deterministic ordering and the sweep work cap. I’ll add a regression with more than 256 older healthy live turns followed by a stale turn and require the stale turn to become eligible within a bounded sweep count before marking this resolved.

PR #1159, opened 2026-09-12
chore(browser): migrate betterwright 2.4.0 to 2.7.3 onto hostTarget leases

kartikkabadi:
## What Changed

- Migrates BetterWright 2.4.0 to 2.7.3 while preserving the unrelated `@pierre/diffs` patch.
- Routes each leased shared Electron session through BetterWright's supplied SOCKS guard proxy before exposing the browser capability; restores the session proxy and drains connections on release, rejecting incompatible concurrent leases.
- Configures BetterWright's Electron network switches before readiness.
- Makes cookie-import navigation, destruction, and timeout interruptions cancel immediately; normal cleanup remains graceful.
- Aligns credential-capture expiry with Synara's native 120-second prompt and adds runtime capture-surface coverage.

## Why

The migration exposed lifecycle and security gaps: interrupted cookie imports could drain instead of canceling, leased shared-session traffic ignored `proxyUrl`, independent Node DNS checks could not prevent Chromium DNS rebinding, session-owned worker requests could bypass `webContentsId` attribution, and capture layers could expire at different times. The session-level BetterWright proxy is the enforcing boundary for navigations, subresources, WebSockets, service workers, and other session-owned requests.

## Verification

- `bun install --frozen-lockfile` passed.
- `bun run fmt:check`, `bun run lint`, `bun run typecheck`, and `bun run windows-runtime:check` passed.
- Browser-automation suite passed: 24 files / 290 tests.
- The full repository run passed all non-web packages; the concurrent web run had 3 timing-sensitive failures in 2 files (`ChatMarkdown.test.tsx` and `Sidebar.import.test.ts`), and those 2 files passed in isolation (62 tests).
- Isolated Synara server/web and desktop development startup completed with debug/WebSocket logging and migrations. The desktop renderer still logs the existing TanStack Router `_nonReactive` error.
- OpenCode 1.18.31 with `opencode/mimo-v2.5-free` completed a non-purchasing Apple catalog/cart attempt; Synara's internal browser connection was unavailable, so live Electron browser automation and live redirect/WebSocket/service-worker/cache/shared-session coverage remain unproven.

## UI Changes

None.

## Checklist

- [x] No UI changes
- [ ] Full repository test run is green


kartikkabadi:
Additional blocking findings from the latest review:

- **Worker rotation keeps an expired proxy** (`betterwrightHostTarget.ts:R45`): `networkGuardLease` is only attached when undefined, so a later `connect({ proxyUrl })` can receive a new worker while the shared Electron session remains routed through the first worker's proxy. Associate the lease with its proxy URL and serialize release/attach so the replacement transport is not vended until the session uses the new proxy.
- **Failed setup can strand the shared proxy** (`betterwrightNetworkGuard.ts:R23-26`): if `setProxy` succeeds but `closeAllConnections()` fails, `activeProxy` is cleared even though the session remains modified. Track setup state, roll back the prior proxy and drain connections before clearing ownership; if rollback fails, retain a failed/owned state that blocks new leases until recovery.
- **Required full suite is still red**: the reported full run has three concurrent web timing failures despite isolated passes. Repository policy requires the broader suite for this lifecycle/security change.
- **Network boundary lacks live Electron coverage** (`betterwrightNetworkGuard.test.ts:15`): current tests mock the enforcement APIs. Add live Electron coverage for guarded traffic, lease rotation, and proxy restoration, including session-owned requests.
- **Capture contract remains locally shadowed** (`browserVaultCapture.ts:101`): the local shim is cast to BetterWright's Playwright context; retain or strengthen real runtime conformance coverage so upstream surface changes cannot silently break Electron capture.

The PR should remain draft and not be considered merge-ready until these are addressed or explicitly accepted as verification gaps.

PR #1162, opened 2026-09-12
fix(server): harden runtime reconciliation recovery

iiiMohammed:
## Summary

- derive runtime-reconciliation staleness from the latest projected message activity, so long-running assistant streams do not hit the 45-minute abandoned-turn guard while preserving the existing per-delta write floor
- include stale `starting`/`running` sessions without an active turn in reconciliation candidates and cover the abandoned-session recovery end to end
- accept stale-generation terminal runtime events when the current binding has no active turn, while continuing to reject events that could clobber a different active turn

## Regression coverage

- planner: recent assistant progress prevents max-age settlement even when session lifecycle timestamps are old
- candidate query: recent message activity excludes a live stream; stale `starting`/`running` sessions with no turn are included; fresh sessions remain excluded
- reconciler layer: a stale `starting` session without a turn is settled to `interrupted`
- provider gate: stale `turn.completed`, `runtime.error`, and `session.exited` are persisted with an empty binding; a stale terminal event for another active turn is ignored

## Verification

- `bun run --cwd apps/server test src/orchestration/threadShellEvents.test.ts src/orchestration/Layers/ProjectionPipeline.test.ts src/orchestration/Layers/ProjectionSnapshotQuery.test.ts src/provider/providerRuntimeReconciliation.test.ts src/provider/Layers/ProviderRuntimeReconciler.test.ts src/provider/Layers/ProviderService.test.ts` (206 tests passed)
- `bun fmt`
- `bun lint`
- `bun typecheck` (7/7 packages)
- isolated dev probe with `SYNARA_PORT_OFFSET=4415`, server port `58290`, web port `10148`, isolated Synara/Codex homes, and `--dry-run` first; a synthetic stale `starting` OpenCode session with no active turn was automatically settled to `interrupted` and emitted `provider.runtime.reconciled`

Out of scope: OpenCode idle/compaction behavior (tracked separately).

### DIFF
--- a/apps/desktop/src/appSnapManager.test.ts
+++ b/apps/desktop/src/appSnapManager.test.ts
@@ -1040,7 +1040,7 @@ describe("AppSnap window picker requests", () => {
       );
       const capture = await capturing;
       expect(capture).toMatchObject({ id: requestId, name: "req-capture.png" });
-      expect(FS.existsSync(capturePath)).toBe(false);
+      await vi.waitFor(() => expect(FS.existsSync(capturePath)).toBe(false));
       expect(await manager.listPendingCaptures()).toHaveLength(1);
       await manager.acknowledgeCapture(capture.id);
       expect(await manager.listPendingCaptures()).toHaveLength(0);
--- a/apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.test.ts
@@ -2374,6 +2374,7 @@ projectionSnapshotLayer("ProjectionSnapshotQuery", (it) => {
       const snapshotQuery = yield* ProjectionSnapshotQuery;
       const sql = yield* SqlClient.SqlClient;
 
+      yield* sql`DELETE FROM projection_thread_messages`;
       yield* sql`DELETE FROM projection_thread_sessions`;
       yield* sql`DELETE FROM projection_turns`;
       yield* sql`DELETE FROM projection_threads`;
@@ -2423,6 +2424,16 @@ projectionSnapshotLayer("ProjectionSnapshotQuery", (it) => {
             'thread-queued-oldest', 'project-runtime-candidates', 'Queued',
             '{"provider":"codex","model":"gpt-5-codex"}', NULL, NULL, 'turn-queued',
             '2026-07-21T00:00:00.000Z', '2026-07-21T00:00:00.000Z', NULL, NULL
+          ),
+          (
+            'thread-running-without-turn', 'project-runtime-candidates', 'Running no turn',
+            '{"provider":"codex","model":"gpt-5-codex"}', NULL, NULL, NULL,
+            '2026-07-20T00:00:00.000Z', '2026-07-20T00:00:00.000Z', NULL, NULL
+          ),
+          (
+            'thread-fresh-queued', 'project-runtime-candidates', 'Fresh queued',
+            '{"provider":"codex","model":"gpt-5-codex"}', NULL, NULL, NULL,
+            '2026-07-23T00:00:00.000Z', '2026-07-23T09:59:00.000Z', NULL, NULL
           )
       `;
       yield* sql`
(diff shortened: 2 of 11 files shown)


iiiMohammed:
**Production validation of both fixes — real stall on 2026-09-16 (v0.8.3, Windows)**

We hit both failure modes this PR targets in a live install today; the evidence maps 1:1 to the two changes.

Thread `ea5b48c7-ad34-41a1-8149-fb5302f42022` (`opencode` / `deepseek-flash`). An external CLI child (`cursor-agent`) hung inside a bash tool with zero output for >80 min. Timeline (server-local time, UTC+3):

1. `21:01:18` — reconciliation settled the stale **running** turn (`provider.runtime_reconciliation.started`, planCount 1). Session became `interrupted`.
2. `21:17:14` — a follow-up user message created a `starting` session with **no active turn** (`activeTurnId: null`; latest turn already `interrupted`).
3. The 45-minute abandonment deadline for that stalled start (`~22:02:14`) passed with **no reconciliation plan for the thread** — nothing was planned at the deadline. This is the candidate gap fixed here: the old predicate admitted only `active_turn_id IS NOT NULL`, `latest_turn.state = 'running'`, or a binding `activeTurnId`, so the planner's guarded abandoned-`starting` branch (`providerRuntimeReconciliation.ts:267-284`) was unreachable in production.
4. `~21:36` — the wedged runtime was aborted; its events were dropped as stale-generation: `provider.session.stale_generation_event_ignored` (event generation `807beef8…` vs current `84c94715…`), repeated for `item.completed` / `thread.token-usage.updated` / `item.updated`, so the stale binding could never settle from runtime evidence. This is the `ProviderService` gate fixed here.
5. The thread stayed `sessionStatus=starting` / `latestTurnState=interrupted` ("Starting OpenCode…") for ~50 minutes and only recovered after a manual `interrupt` + new message. With this PR it would have self-settled at the 45-minute mark instead.

Also seen in the same window (likely separate, flagging for context only): a burst of `orchestration dispatch timed out before command started` (45000 ms) — including one `thread.turn.start` that never started — and repeated `CheckpointStore.captureCheckpoint` failures (`git add -A` output >1 MB truncated) on a large repo.

Residual questions from the diff:

- A never-bound start still waits the full `RUNTIME_RECONCILIATION_MAX_TURN_AGE_MS` (45 min). Since `thread.turn-start-requested` delivery is already bounded at 120 s, a dedicated shorter bound for `starting`-without-turn sessions may be worth considering.
- After the abandoned-start settle, the pending user message is not re-dispatched (the user resends). Out of scope is fine; just noting it as a follow-up decision.

Thanks for the isolated-instance probe and the regression coverage — both changes are needed.

iiiMohammed:
Follow-up pushed in c21a44610: sessions stranded in `starting`/`running` with no turn now settle on a dedicated stalled-start bound (`RUNTIME_RECONCILIATION_MAX_STALLED_START_AGE_MS`, 10 minutes, overridable through `ProviderRuntimeReconcilerLiveOptions`) instead of waiting out the 45-minute max turn age.

- The null-turn settle branch uses the shorter clock; the pump-health and runtime-journal guards still hold settlement off and keep the max turn age as their escape hatch, so ingestion lag can still prevent a premature settle.
- Coverage: planner cases (stalled start settles, fresh start stays, lagging journal holds) plus a reconciler-layer test proving an injected stalled-start clock settles a one-minute-old start.
- Locally green: focused provider reconciliation tests, `bun fmt`, `bun lint`, `bun typecheck` (7/7 packages).

iiiMohammed:
Re the two Devin findings from the earlier review — both were already addressed in 5ee285291 (the merge-repair commit pushed after that review) and have dedicated coverage:

**Bug 1 — stale events retire replacement sessions:** `staleTerminalSettlesThread` and `canSettleBinding` now require `binding.lifecycleGeneration === event.lifecycleGeneration` (or the event naming the binding's active turn); an idle replacement binding is no longer retired by a stale exit/error. Coverage in `ProviderService.test.ts` ("serializes lifecycle mutations and persists a fresh generation per start") emits a stale `session.exited` from the old generation and then a current-generation `session.state.changed`, asserting the newer binding keeps its generation, status, and active turn.

**Bug 2 — concurrent streams appear abandoned:** `observed_at` is now scoped to messages of in-flight turns (`sessions.active_turn_id` / runtime binding `activeTurnId` / `latest_turn.state = 'running'`) instead of the globally latest message. Coverage in `ProjectionSnapshotQuery.test.ts` ("lists only stale active thread ids for reconciliation, oldest first") seeds a still-streaming turn alongside a newer settled turn with a higher message sequence and asserts the streaming thread stays out of the candidate list.

Re-verified on the current head (`c21a44610`): `bun run --cwd apps/server test src/provider/Layers/ProviderService.test.ts src/orchestration/Layers/ProjectionSnapshotQuery.test.ts` → 135 passed.

PR #1227, opened 2026-09-15
Ambient in-chat computer preview plus foreground focus restore

kartikkabadi:
Draft: ambient Codex-style live preview of the driven desktop inside the chat, plus focus hygiene for foreground input.

Stack (14 atoms, oldest first):
- Pop reveal motion for disclosure surfaces
- Frame tap prime after launch_app (live from cold start)
- Tap reports decoded frame size; only the driving thread draws
- Stills stream only clears a canvas it owns
- Floating browser chrome convergence
- Preview layout footprint in the store, ambient rail slot under the Environment card
- Preview size setting (compact default) with shared caps plus rail budget
- Expand keeps the session live (closing the dock pane brings the card back)
- Chromeless dynamic replica card with glass finish
- Preview mounted in the environment rail with gutter inset
- Foreground calls restore the user's frontmost window afterwards

Verified: web typecheck plus lint plus Vitest (unit plus browser), server computer plus gateway suites. Packaged 0.8.4 artifact tested locally.


kartikkabadi:
## Computer Use implementation status

This branch now contains the completed implementation work from the macOS Computer Use investigation. The packaged runtime proofs are intentionally reported separately from source/unit verification so unverified behavior is not overstated.

### Implemented

- Exact `(pid, window_id)` target admission with WindowServer ownership, positive geometry, visibility, Space membership, and Accessibility ancestry checks.
- Focus-neutral semantic text insertion through the exact writable Accessibility element. Background semantic input does not use process-wide HID or transient foreground activation.
- Concurrent native semantic leases and server-side scoped scheduling:
  - different exact targets may overlap;
  - the same target is not admitted concurrently;
  - pointer and generic keyboard operations remain exclusive;
  - cancellation and shutdown drain active controllers.
- Multi-Space safety contract:
  - active-Space routing for pointer, generic keyboard, scroll, drag, activation, and foreground actions;
  - stable-membership routing only for exact retained semantic text targets;
  - revalidation of identity, ownership, geometry, membership, and AX ancestry;
  - fail-closed refusal for ambiguous or stale targets;
  - off-Space screenshot freshness is explicitly unverified rather than presented as live grounding.
- Per-thread browser/Computer tool guidance on first use and every tenth subsequent use, with bounded LRU storage.
- Browser guidance for resumable, structured long-history extraction preserving text, links, media, ordering, checkpoints, and honest boundaries.
- Cua capability audit covering the pinned native surface and deliberately constrained raw operations.
- Removed and disabled the misleading right-sidebar Computer pane while preserving the ambient in-chat live preview and backend Computer tools.
- Added a foreground-only composer-focus guard. A pending focus request now waits for a real window-focus event instead of activating Synara after a provider turn completes in another app or Space.
- Rebuilt unsigned DMGs from the final staged app and then regenerated the update ZIP, preventing DMG/ZIP payload skew.

### Source verification

- `bun run fmt:check`: passed.
- `bun run lint`: passed with 0 errors and 630 repository warnings.
- `bun run typecheck`: passed across all 7 packages.
- `bun run windows-runtime:check`: passed across 249 application source files.
- Native `rustfmt --edition 2021 --check`: passed.
- `cargo check -p cua-driver`: passed with one existing dead-code warning.
- React Doctor changed-scope result: 65/100, unchanged from the prior recorded score; diagnostics remain in the existing changed surface.
- Full `bun run test` was run twice:
  - the first parallel verification run hit three 5-second desktop timing failures; all three affected files then passed focused (83/83);
  - the clean full rerun reached one 5-second `ChatMarkdown` import timeout; that file then passed focused (61/61).
  - Because the aggregate command did not finish green, this is reported as a verification gap rather than a full-suite pass.

### Fresh packaged artifacts

Built from the working tree committed as `bb7eb421c`:

```text
b0505ee9432f465b7fac02b9df6b725bea984e757443d80ca5455058feb632bd  Synara-0.8.4-arm64.dmg
8ddf0d43030d8bf72f62c5b4d78255b3e99e1216f369f60141f55da96a3f5010  Synara-0.8.4-arm64.zip
```

The DMG checksum verifies with `hdiutil verify`, includes the `/Applications` symlink, and contains:

```text
bundle identifier: com.emanueledipietro.synara
version: 0.8.4
Cua Driver: 0.24.0
upstream revision: 4b3396d9fe4bd3cf723b0eb8db83c18a8764b520
Synara native revision: 14
native patch SHA-256: 3375ac1c2fe8266672ac1735b8514a8f2f692a64b19d415630745478d0c19a32
arm64 driver SHA-256: b9132545620db2bb5501f74194d66cc47aa5d9bbb10663e1c755962b853ff0a9
```

The artifact is unsigned, as expected for the local unsigned build path. It is **not Apple-signed or notarized**. Strict code-sign verification therefore does not pass until an installed local copy is recursively ad-hoc signed or the release signing pipeline signs it.

### Live-testing findings and remaining gaps

The previous revision-14 packaged certification is not valid success evidence:

- Safari was initially key on Desktop 1, but Synara on Desktop 2 became frontmost without explicit activation.
- The sampler recorded Synara as frontmost with `AXMain=true` for all 120 samples.
- Human input entered Synara's composer.
- SQLite inspection showed only `computer_get_state` events for that provider turn and no persisted `computer_type_text` event.
- The TextEdit marker attributed to the turn was already present before the alleged mutation.

The new composer-focus guard addresses the observed activation path, but the newly rebuilt package has **not** received a fresh UI-driven certification run.

Still unproven:

- Three simultaneous provider-backed Synara threads typing into three real background applications while a fourth foreground app remains continuously usable.
- Fresh persisted `computer_type_text` lifecycle events plus before/after target readback.
- Retained exact semantic input into an off-Space application.
- Screenshot freshness refusal and active-only pointer/generic-keyboard refusal in the packaged app.
- Identity, Space-membership, hidden-window, and minimized-window interruption/refusal behavior in the packaged app.
- Fresh-agent extraction of a genuinely long, complex, mixed-media history.
- The canonical macOS Lume harness, which requires a logged-in maintainer VM with SIP disabled, the signing keychain, GUI access, and TCC grants.

The implementation and artifacts are now synchronized on the PR branch, but this PR should remain draft until those packaged runtime proofs are completed.

[Written by Devin](https://app.devin.ai/sessions/d457b1a48fe845aa845df2a8255124e4)

PR #1267, opened 2026-09-18
fix(process-tree): verify pid ownership before every signal

kartikkabadi:
## Summary

Teardown could signal stale or recycled PIDs and re-walked live process trees at signal time. A recycled root could target a replacement process, and a fake handle (`pid: 1`) became `kill(-1)` — broadcasting SIGTERM to every user process.

This PR makes every signal require proof first:

- **PID guards** — `pid <= 1` and malformed PIDs are never signaled or walked (`signalPid`, `capture`, `captureProcessTree`, `signalOwnedChildProcess`).
- **Spawn-time identity** — spawn sites record each child's live OS identity (`ppid`, `pgid`, `lstart`/CIM CreationDate) via `noteSpawnedProcess(pid)` into a module-level registry.
- **Verified root gate** — `signal()` probes the live row and requires `ppid === us` plus a start-time (or command) match against the registry or captured root. An unverifiable root signals nothing — the captured tree is treated as foreign.
- **Verified descendants** — captured descendants are signaled only when their live start-time identity still matches; POSIX never re-walks a live tree at signal time.
- **Windows** — `taskkill /T` still reaches `.cmd`/`cmd.exe` descendants, but only after the root verifies via a single-row CIM probe (`readWindowsProcessRow`).
- **Broadcast guard** — `process.kill` refuses `-1`, `0`, `1`, and negative-pid group signals whose group leader is not our live child (`pgid === pid`). This covers the dependency's `kill(-pgid)` scope teardown without a package patch.

Fail closed throughout: a failed probe, missing row, or mismatched identity signals nothing.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bunx vitest` focused suites — 644 passed (platform, terminal, device, providerUsage, provider, processRunner)
- [x] `bunx oxlint` — 0 errors; `oxfmt` clean on touched files
- [x] `scripts/check-windows-runtime-boundary.ts` — verified
- [x] Regression tests: `kill(-1|0|1)` refused; stale-root `startedAt` mismatch signals nothing; verified root+descendants signaled
- [ ] Windows runtime verification (CIM probe + `taskkill` path) — no Windows host available; compile-verified only

## Known limits

- POSIX `lstart` is second-resolution: a same-pid recycle within one second is the residual ambiguity window.
- Verify-then-signal is not atomic on POSIX (no fd-held process handle); the window is probe→kill.


kartikkabadi:
Addressed the review findings:

**Windows child trees survive cancellation** — fixed in `b08ad2c72`. After the root passes identity verification, `signal()` on win32 now retires it through the native `taskkill /pid /T /F` traversal (via the `signalTree` dep) instead of a per-PID signal, so a `.cmd`/`cmd.exe` shim's real worker is terminated too. Only fires for a verified root on Windows — POSIX is unchanged, and an unverified root still signals nothing. Covered by 5 new tests (verified root → taskkill; foreign root → nothing; `includeRootTree:false` → per-descendant path preserved for the teardown escalation).

**Verification and signaling remain non-atomic** — acknowledged. POSIX has no atomic verify-and-signal primitive (no pidfd on macOS), so a probe→signal TOCTOU window is inherent. The residual is bounded to the interval between the `ps` snapshot and `process.kill` (microseconds), versus the previous design where the stale window was unbounded (captured earlier, signaled later, or live-tree expansion from a recycled root). No further mitigation available without a platform primitive.

**Synchronous probes delay server teardown** — acknowledged, bounded. Each `signal()` call makes exactly one `ps` probe covering the root + all captured descendants (not one per PID), and probes only run on teardown paths, never in hot loops. The Windows PowerShell probe is heavier (up to the 3s timeout); a future improvement could route it through the existing persistent snapshot observer instead of a one-shot spawn — noted, not blocking.

**Reparented children skip direct teardown** — acknowledged, rare edge. A direct child keeps `ppid === our pid` while we are alive; it can only be reparented by deliberately daemonizing (double-fork/`setsid`), at which point skipping the signal is arguably correct — it detached itself. If a provider ever intentionally daemonizes, the teardown's captured-descendant path still reaches its children.

Verification after the fix: `processTreeController.test.ts` 48/48, `processTreeKiller.test.ts` 8/8, plus adjacent suites (Manager, supervisedProcessTeardown platform+provider, codexResetCredits) 100/100; typecheck, lint, `windows-runtime:check` all green.

kartikkabadi:
Round 2 addressed — the pid-reuse identity findings were real and are now fixed:

**Recycled sibling receives teardown signal** — fixed. Root signaling now requires identity *evidence*, not just `ppid === ownerPid`: a captured root row that still matches, or a caller-supplied `expectedRootCommand` matching the live row. POSIX captures and probes now include `lstart` (process start time — the POSIX analogue of Windows `CreationDate`), so a pid recycled by a *sibling* child under the same parent is rejected by start-time mismatch. `supervisedProcessTeardown`, `opencodeRuntime`, and `AcpSessionRuntime` pass their spawn command through; `helperClient.dispose` and the dep patch's `isOwnedChildPid`/`isOwnedProcessGroup` now compare the live command too. Where command evidence isn't available (`Manager`'s pty path, `process.title` rewrites) the check stays ppid-only — documented trade-off.

**Reparented descendants escape teardown** — fixed. `isSameLiveProcess` no longer requires the live PPID to equal the captured PPID when a start-time identity exists: command + `lstart` match is sufficient, so an orphan reparented to pid 1 is still correctly identified and signaled. The ppid check remains as fallback only when no start time exists on either side.

**`expectedRootCommand` bypassed on the verified path** — fixed, exactly the suggested change.

**Patch reparented groups skip cleanup** — already handled. A group leader that "stops being a direct child" has exited (a live process cannot change parents while we are alive) → the leader row is absent → leaderless groups are signaled. Only a *live foreign* leader blocks the group kill, which is the intended protection.

**Non-atomic verify→signal** — acknowledged; inherent on POSIX, bounded to microseconds. No atomic verify-and-signal primitive exists on macOS.

**Synchronous probes** — acknowledged; one `ps` call per `signal()` invocation, teardown paths only. Windows probe future improvement noted (route through the persistent observer).

**Windows runtime unverified** — acknowledged in the PR body. Source-level tests + injected platform seams only; needs a real Windows host to confirm end-to-end.

Verification on the pushed commits: 192/192 tests across 8 affected files (controller 55, Manager 56, helperClient 12, teardown 10+16, killer 8, spawner-guard 3, opencodeRuntime 32), typecheck clean, lint 0 errors, `windows-runtime:check` pass. The dep patch was verified to apply cleanly and reproduce the installed copy.

kartikkabadi:
Round 3 addressed:

**Stale root kills sibling tree (R413-414)** — fixed in `1350d0416`, exactly per the recommendation: `expectedRootCommand` is now a hard conjunctive gate whenever provided; identity evidence (matching captured root OR the expected command) stays disjunctive. A sibling captured after pre-capture recycling can no longer satisfy the gate.

**Windows evidence gap (found while fixing the above)** — the tightened gate exposed a real regression: `PtyProcess` carried no `spawnargs`, so on win32 (empty sync capture) terminal teardown had no identity evidence and would have failed closed entirely. `a3435eeaa` adds `spawnargs` to the `PtyProcess` contract, populated at `NodePTY`/`BunPTY` spawn, threaded through `killProcessWithEscalation` into both the ownership probe and the tree signal.

**Script children survive scope cleanup (patch R203)** — fixed in `08ff336df`, per the recommended approach: the patched spawner now captures `{ppid, lstart, command}` synchronously right after `spawn()` and verifies `pid + start-time + ppid` at signal time — command-string matching is gone entirely. Process-group checks use the same identity (leader ppid + start time; leaderless groups still accepted). Windows got the equivalent `CreationDate` identity via CIM, so the guard is platform-unified. New test proves a `#!/bin/sh` script child IS terminated on scope close — it failed under the previous guard.

**Recycled sibling (R395-399)** — same finding as round 2, already addressed there; `lstart` now also distinguishes same-command siblings. Residual: POSIX `lstart` is second-resolution, so two identical-command children spawned in the same second that collide on a recycled pid remain indistinguishable — no finer portable identity exists on macOS via `ps`.

**Non-atomic verify→signal** — acknowledged (third occurrence); inherent, bounded to microseconds.

**Synchronous probes** — acknowledged; one `ps` per `signal()`/spawn, teardown and spawn paths only. The dep patch now adds one `ps` per Effect spawn — same bounded trade.

**Windows runtime unverified** — acknowledged in the PR body; needs a real Windows host. The patch's win32 identity path follows the existing CIM pattern but is untested on hardware.

Verification: 153/153 tests across the 6 affected files (controller 61, Manager 58, helperClient 12, teardown 10, killer 8, spawner-guard 4 incl. the shebang case); typecheck clean; lint 0 errors; `windows-runtime:check` pass; dep patch verified to `git apply` cleanly and reproduce the installed package byte-for-byte.

kartikkabadi:
Round 4 addressed — this round's findings were all real, and the fixes converge on the architecture the earlier rounds were pointing at:

**Transformed providers evade teardown (R417-419)** — fixed by replacing command-string evidence entirely. Command lines were never a valid identity: shebang scripts (`/bin/sh /path/tool`), Windows `makeEffectProcessCommand` planning (`.cmd`→`cmd.exe`, WSL), `sandbox-exec` exec-in-place, and `process.title` all legitimately change the live argv. The controller now exports `captureProcessIdentity(pid)` which records `{pid, ppid, startedAt, command}` from the live process table **immediately after spawn**, and every kill path accepts it: `expectedRootIdentity` on `signal`/`signalProcessTree`/`supervisedProcessTeardown`, `expectedIdentity` on `isOwnedDirectChildPid`, `processIdentity` on `signalOwnedChildProcess`/`PtyProcess`/`ProcessExitHandle`. Gate precedence: a start-time match alone authorizes (command cannot veto a verified identity); without a start time the *observed* captured command outranks the logical one; with neither, the previous weak-evidence rules apply unchanged. Wired at every spawn site: `opencodeRuntime`, `AcpSessionRuntime`, `AntigravityAdapter`, `ClaudeAdapter`, `PiAdapter`, `processRunner`, `codexAppServerManager`, `codexResetCredits`, `IosSimulatorBackend`, `NodePTY`/`BunPTY` (replacing the spawnargs approach), `helperClient`. Three raw `SIGKILL` fallbacks were also rerouted through the guarded `signalOwnedChildProcess` path.

**Process spawns freeze the server (patch R14-16)** — fixed. The per-spawn probe is now **async**: `execFile` on the raw `node:child_process` module after the child's `spawn` event (deliberately not `NodeChildProcess.spawn`, so the probe can never re-enter the instrumented spawner). No synchronous subprocess on the spawn path at all. Kill-time probes stay synchronous (they only run during teardown). If finalization runs before the async probe lands, the map has no entry → fail closed, kills skipped — bounded window, documented in the patch.

**Delayed exits erase new identities (patch R22-24)** — fixed. The `exit`/`error` listeners capture the stored identity object and delete only when `spawnedIdentities.get(pid) === spawnedIdentity`, so a recycled pid's newer record survives a stale callback. Applied to both `dist` and `src` hunks.

**Patch whitespace** — fixed. The patch was regenerated from the pristine tarball (fetched from the resolved pkg.pr.new URL); `git apply` on a fresh pre-image reproduces the installed `dist` + `src` byte-for-byte, and the file contains no trailing whitespace.

**PR scope (~2.7k lines)** — acknowledged. ~55% is test code; the patch necessarily mirrors `dist` and `src` hunks. If you'd prefer, I can split the `patches/` change into a separate PR — it is self-contained. The rest is one semantic change (spawn-time identity verification) that only makes sense end-to-end.

**Non-atomic verify→signal** — acknowledged (fourth occurrence). No POSIX primitive exists; window is bounded to microseconds between probe and signal.

**Sync probes at teardown** — partially improved this round (spawn path is now fully async). Teardown-path `ps` probes remain sync — one probe per `signal()` call, only on teardown paths. Routing them through a persistent observer remains a possible future optimization.

**Windows runtime unverified** — acknowledged in the PR body; needs a real Windows host.

Verification: 237/237 tests across 9 affected files (controller 74, Manager 58, helperClient 14, teardown 11+16, killer 8, spawner-guard 4, opencodeRuntime 32, codexResetCredits 19); typecheck clean; lint 0 errors; `windows-runtime:check` pass; patch verified byte-for-byte against pristine.

Residual (documented in code): POSIX `lstart` is second-resolution — two same-command siblings colliding on a recycled pid within the same second are indistinguishable; no finer portable identity exists on macOS via `ps`.

kartikkabadi:
Rewrote this PR from 3,747 to 438 lines (17 files, one commit) per the review-scope finding — the previous revision threaded an identity object through every call site and patched the Effect dependency (465 lines of patch alone).

Same safety properties, much smaller mechanism:

- **One registry, zero threading**: `noteSpawnedProcess(pid)` at spawn sites records live OS identity into a module-level map. `signal()` consults it internally — no signature changes, no `expectedRootIdentity` plumbing. 12 one-line call sites; every other file reverted to upstream.
- **`process.kill` broadcast guard replaces the dep patch**: refuses `-1`/`0`/`1` and negative-pid group signals whose leader isn't our live child (`ppid === us && pgid === pid`). Covers the spawner's `kill(-pgid)` and any future caller, no patch file to drift.
- **Controller keeps upstream shape**: `pid <= 1` guards, `lstart` captured with the tree, root signaled only after live `ppid` + start-time identity match, descendants signaled only on identity match. POSIX never re-walks a live tree; Windows `taskkill /T` runs only after the root verifies via a new single-row CIM probe (~60 lines).
- **Tests**: ~70 lines covering broadcast refusal, stale-root refusal, and verified signal — the large parameterized suites are gone.

Net: +438/-76 vs the previous +3,747. All earlier findings remain addressed — fail closed on missing evidence, recycled-PID rejection via start-time identity, reparented descendants still match via `lstart`, and the group-kill path cannot turn a fake `pid: 1` into `kill(-1)`.

kartikkabadi:
Post-rewrite self-review found and fixed two real bugs in the gate:

- **Dead-root leak**: a root that exited between capture and signal returned early, leaking every verified descendant — the common teardown path (root exits, then stragglers are cleaned up). Now a probe *failure* vetoes everything, but a *dead* root still signals individually-verified descendants when the registry proves the capture ran while our child held the pid (captured root `startedAt` === spawn identity).
- **Sibling-recycle acceptance**: a sibling that recycled the pid before capture satisfied both the captured-root and live-row checks simultaneously. Now when a spawn-time identity exists, the captured root's `startedAt` must match *it* — capture↔live consistency alone doesn't prove the capture saw our child.
- Also: `signalOwnedChildProcess` treats a recorded spawn identity as authoritative (ppid can't override it), and `readSpawnIdentity` is injectable so both paths are unit-tested (+2 regression tests).

Net diff is now +498/-76.

kartikkabadi:
Superseded by #1270 (codex/safe-process-signals), which merged the same guard work. Closing.

PR #1276, opened 2026-09-20
feat: passively return delegated gateway task results

### LINKED ISSUE #1275: [Feature]: Agent Gateway: automatically return delegated task results to the parent thread
### Before submitting

### Area

apps/desktop

### Problem or use case

Synara's Agent Gateway allows a task to delegate work to another Synara thread with `synara_create_thread`.

The problem is that there is currently no durable completion callback from the delegated child thread back to the parent thread.

My typical workflow is:

1. I am actively working with a parent agent, for example Codex/Astra.
2. The parent delegates implementation or review to Claude Code/OpenCode in a separate Synara thread.
3. I want to continue talking and working with the parent immediately while the child runs in parallel.
4. When the child finishes, I want its result to automatically arrive back in the parent thread.

Today I effectively have to choose between:

```text
Parent
  ↓
create child
  ↓
wait_for_threads
  ↓
wait again
  ↓
read_thread
  ↓
continue
```

which keeps the coordinating turn occupied,

or:

```text
Parent
  ↓
create child
  ↓
parent finishes its turn

...child works in background...

User later:
"Did the child finish? Read its result."
```

which requires manual intervention.

The missing primitive is:

```text
delegate child
→ parent remains usable
→ child finishes
→ result is automatically delivered to parent
```

### Proposed solution

Make child completion routing a first-class, server-owned property of delegated Synara tasks.

For example:

```text
synara_create_thread({
  ...,
  completionTarget: {
    threadId: "<parent-thread-id>",
    mode: "notify"
  }
})
```

or, because Synara already knows which thread called `synara_create_thread`, potentially something as simple as:

```text
notifyParentOnComplete: true
```

When the child reaches a terminal state, Synara would automatically deliver a structured completion message/event to the parent thread.

For example:

```text
Child completed

thread: <child-thread-id>
status: completed
provider: Claude Code
model: Opus
result: <final child result / summary>
```

Failures should be routed as well:

```text
Child failed

thread: <child-thread-id>
status: failed
reason: ...
```

If the parent currently has an active turn, the completion message could simply be queued using the existing follow-up behavior instead of interrupting the active turn.

If the parent is idle, the completion result would be waiting there the next time the parent runs.

An optional future mode could also explicitly continue/wake the parent, but automatic continuation is not necessary for the initial feature.

### Why this matters

This would make Agent Gateway much more useful as an actual orchestration layer rather than only a task creation/control API.

It would enable workflows such as:

```text
Parent / coordinator
       │
       ├── OpenCode implementation
       │          ↓
       │     completion returned
       │
       ├── Claude review
       │          ↓
       │     completion returned
       │
       └── OpenCode fixes
                  ↓
             completion returned
```

while still allowing me to continue working interactively with the parent during long-running child tasks.

The main benefit is that the parent does not need to remain blocked in repeated `synara_wait_for_threads` calls just to receive a result that Synara already knows has completed.

It also makes multi-provider orchestration substantially more reliable because task completion becomes infrastructure-owned rather than dependent on model behavior.

### Smallest useful scope

The smallest version that would solve the problem is:

When thread A creates thread B through Agent Gateway, Synara records A as the creator/parent.

When B reaches any terminal state, Synara automatically appends a completion message to A containing:

```text
- child thread ID
- terminal status
- final child response/result
- failure reason, if applicable
```

No automatic parent wake-up, continuation, workflow engine, retries, or DAG orchestration is required initially.

Simply delivering the result back to the creating thread would solve most of the problem.

### `synara_wait_for_threads`

The parent can wait for the child and then read its result.

This works, but it means the parent has to remain occupied waiting for potentially long-running work.

The bounded wait also needs to be repeated for long tasks.

### Manually checking later

The parent can create the child, finish its turn, and I can later ask:

```text
Check whether thread X finished and read its result.
```

This works but requires manual bookkeeping and user intervention.

### Ask the child to call `synara_send_message`

A workaround is to pass the parent thread ID into the child prompt:

```text
Parent thread ID: <id>

When your task is finished, use synara_send_message
to send your final result back to the parent thread.
```

This approximates the desired behavior and is probably what I will use for now.

However, it depends on the model remembering to make the final tool call.

It also does not reliably cover:

```text
- provider crashes
- context exhaustion
- interrupted tasks
- permission/tool errors
- process failures
- other terminal states where the model never gets to execute the callback
```

Synara itself already knows when the task enters a terminal state, so completion delivery seems better implemented at the server/orchestration layer.

### Risks or tradeoffs

A few edge cases would need clear behavior:

### Parent currently has an active turn

The completion should probably be queued rather than injected unexpectedly.

### Parent thread was deleted or archived

The result could remain available only in the child, or Synara could mark delivery as unavailable.

### Child failure

Failures should still generate a completion notification with status and error information.

### Recursive delegation

If children create children, completion should only route to the explicitly recorded parent/creator unless configured otherwise.

### Notification volume

Large fan-out workflows could produce many completion messages. This could potentially be handled later with batching, but it should not block the basic implementation.

### Permissions

Completion delivery should respect existing project/thread permissions and should not create a privilege escalation path.

### Examples or references

Current workflow:

```text
Astra parent
    │
    ├── synara_create_thread(...)
    │
    ▼
Claude Opus child
    │
    │ works for several minutes
    ▼
DONE

Astra does not automatically receive the result.

The user must either:
- keep Astra waiting with synara_wait_for_threads, or
- manually ask Astra later to inspect the child.
```

Desired workflow:

```text
Astra parent
    │
    ├── delegate → Claude Opus
    │
    ├── user continues working with Astra
    │
    │
Claude Opus works in parallel
    │
    ▼
DONE
    │
    └── Synara automatically posts result to Astra thread
```

I like that delegated threads remain independent, visible tasks with their own lifecycle.

I am not suggesting changing that model.

I am specifically requesting an optional durable completion route from a delegated child back to the thread that created it.

### Contribution

- [x] I would be open to helping implement this.

Emanuele-web04:
When an agent delegates a task, its creator currently has to poll or manually read the child to learn the result. Add `notifyCreatorOnComplete: true` to single and batch gateway creation: Synara durably returns one bounded, attributed result for the initial run to the authenticated creator, while the child remains standalone.

The result uses the existing activity UI and becomes untrusted reference context on a later human-started provider turn. Delivery never starts, queues, or steers parent work and preserves human-message recency. The durable outbox binds the deterministic initial message to its provider turn, waits for runtime ingestion acknowledgement, requires a real completion event for success, and uses deterministic activity receipts for replay. Context assignments share provider-command settlement: rejected sends release them, retries retain their assignment, and uncertain sends remain held behind reconciliation. Archived/deleted creators receive no activity; delivery is recorded unavailable in the child.

Goals are explicitly unsupported: an initial run that used a goal reports that limitation instead of declaring an intermediate goal turn complete. Later child turns do not notify again. Summaries and errors are capped at 2,000 characters with truncation flags; each result references `synara_read_thread`. Context is capped at 16,000 characters per send. No new dependencies or UI components.

Validation:
- `bun run fmt:check`, `bun run lint` (0 errors; existing warnings), `bun run typecheck`, `bun run migrations:check`: passed.
- Four focused server files: 475 tests passed. Final completion/passive regression pass after the last refinements: 28 passed.
- Full `bun run test` passed with the repository-pinned Node 24.13.1: all 7 workspace tasks succeeded; server 5,456 passed / 20 skipped, web 4,766 passed / 3 skipped, desktop 870 passed / 6 skipped. Subsequent human-origin and replay refinements passed their focused regressions and all static checks.
- Migration/upgrade regressions: 41 passed. Final replay regression file: 21 passed, including the real orchestration command fingerprint check.
- Earlier full runs exposed sandbox socket `EPERM` and installed Node 26 localStorage fixture failures; the pinned-node run permits fixture sockets without changing global configuration.
- No paid/live provider or signed-desktop validation. Existing activity rendering is reused; no new transcript follow behavior.

Closes #1275

Cursor found one replay fingerprint issue on the initial head; fixed in `1a604f64f` by persisting the activity timestamp with the result and checking the real command fingerprint in the restart/retry regression. Cursor reviewed that exact head with no issues found, and its initial CI run passed. The PR is now ready for review. The fresh full CI rerun (run 35525416352, attempt 2) passed all 18 CI jobs on `1a604f64f`; Cursor remains clean on that exact head. No merge has been performed.

---

> [!NOTE]
> **Medium Risk**
> Touches orchestration provider sends, durable outbox semantics, and activity commands; mistakes could leak wrong child output into parent prompts or drop/replay completions incorrectly.
> 
> **Overview**
> Adds **passive completion delivery** for delegated gateway tasks via optional **`notifyCreatorOnComplete`** on `synara_create_thread` / `synara_create_threads` (authenticated provider sessions only; external MCP is denied).
> 
> When creation finishes, opt-in child threads are recorded in a new **`agent_gateway_completions`** outbox (migration 108). A **~1s background scanner** (`deliverGatewayCompletions`) waits for runtime ingestion, pins the **initial message/run** (not later turns), builds a capped summary/error payload, and appends **`synara.task.completed`** activity on the creator—or records **unavailable** delivery on the child if the creator is archived/deleted. Goals on the initial run are reported as unsupported.
> 
> **Provider sends** gain bounded (**16k**) untrusted completion context on **human-started** turns only; assignment is tied to provider-command settlement (rejected sends release context, accepted sends consume it). **`thread.activity.append`** can set **`requireUnarchived`** so creator notifications skip archived threads while still allowing child-side unavailable activities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a604f64f6a13f43bbf54a5bcdca2d3c707f2c59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Emanuele-web04:
bugbot run verbose=true

Emanuele-web04:
bugbot run verbose=true

